PR #3, opened 2019-12-11
Update/add some random things

SPGoding:
image
sad ;(

- - -
This bot is really awesome. I thought that I could add some doc for those classes, and then found that the structure is so self-descriptive that there's no need to add extra comments. So I just fix some random things and add a script for eslint ;)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "A work-in-progress bot used to link to Mojira bug reports directly from Discord.",
   "main": "index.js",
   "scripts": {
+    "lint": "eslint **/*.ts --fix",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "author": "violine1101",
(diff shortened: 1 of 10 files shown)


violine1101:
You should have write access to the repository (except `master` branch) now. Sorry, I forgot to configure that.

There were still a few things that I only got to implement/change halfway. I'll review and merge the PR tomorrow. Feel free to add more stuff until then. Thanks for helping out! :)

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "A work-in-progress bot used to link to Mojira bug reports directly from Discord.",
   "main": "index.js",
   "scripts": {
+    "lint": "eslint **/*.ts --fix",
```
violine1101: This command doesn't appear to work properly, presumably because this project uses TypeScript ESLint, not vanilla ESLint.

SPGoding: But it works on my machine 🤔

bemoty: Also works for me

violine1101: Ah, I tried to run that command directly on the command line instead of using the `npm run lint` command, which obviously didn't work.

PR #4, opened 2020-01-11
Add jira filter feed feature

SPGoding:


src/tasks/TaskScheduler.ts
```diff
@@ -0,0 +1,7 @@
+import Task from './Task';
+
+export default class TaskScheduler {
+	public static add( task: Task, interval: number ): void {
+		setInterval( task.run, interval );
```
violine1101: Current issue with this implementation: The task is not canceled and the interval still continues even after the bot has been shutdown using `!jira shutdown`. Only relevant for me, but I plan to add a restart command to the bot at some point, which would make this even more messy.

src/tasks/FilterFeedTask.ts
```diff
@@ -0,0 +1,83 @@
...
+			host: 'bugs.mojang.com',
+			strictSSL: true,
+		} );
+	}
+
+	public async run(): Promise<void> {
+		let upcomingTickets: string[];
+
+		try {
+			const searchResults = await this.jira.search.search( {
```
violine1101: Attention: `this` doesn't refer to `FilterFeedTask` here, since you run this method using `setInterval`! JavaScript/TypeScript is tricky in this regard.

```
[2020-01-12T14:18:49.575] [ERROR] FilterFeed - TypeError: Cannot read property 'search' of undefined
    at Timeout.<anonymous> (...\mojira-discord-bot\src\tasks\FilterFeedTask.ts:37:42)
    at Generator.next (<anonymous>)
    at ...\mojira-discord-bot\src\tasks\FilterFeedTask.ts:8:71
    at new Promise (<anonymous>)
    at __awaiter (...\mojira-discord-bot\src\tasks\FilterFeedTask.ts:4:12)
    at Timeout.run [as _onTimeout] (...\mojira-discord-bot\src\tasks\FilterFeedTask.ts:30:16)
    at listOnTimeout (internal/timers.js:531:17)
    at processTimers (internal/timers.js:475:7)
```

src/tasks/FilterFeedTask.ts
```diff
@@ -0,0 +1,83 @@
...
+
+	private client: Client;
+	private jira: JiraClient;
+	private channel: string;
+	private jql: string;
+	private title: string;
+
+	private knownTickets: Set<string>;
+
+	constructor( { channel, jql, title }: FilterFeedConfig, client: Client ) {
```
violine1101: I tried to cap off `MojiraBot.client` as far as possible. Would be good if it was not a necessary argument. Perhaps the `Channel` object should be passed over instead?

I mean it works like this, but I just try to avoid it in order to not give smaller classes power over the entire bot. That's not the point of modularization.

src/tasks/FilterFeedTask.ts
```diff
@@ -0,0 +1,83 @@
...
+		if ( this.knownTickets ) {
+			const unknownTickets = upcomingTickets.filter( key => !this.knownTickets.has( key ) );
+
+			if ( unknownTickets.length > 0 ) {
+				try {
+					const embed = await MentionRegistry.getMention( unknownTickets ).getEmbed();
+					embed
+						.setTitle(
+							this.title
+								.replace( /\{\{num\}\}/g, unknownTickets.length.toString() )
```
violine1101: This looks a bit ugly, perhaps there's another way to input parameters into a string similar to `printf`.

src/tasks/FilterFeedTask.ts
```diff
@@ -0,0 +1,83 @@
...
+						throw `Expected ${this.channel} to be a TextChannel`;
+					}
+				}
+				catch ( err ) {
+					FilterFeedTask.logger.error( err );
+					return;
+				}
+			}
+		}
+		upcomingTickets.forEach( key => this.knownTickets.add( key ) );
```
violine1101: Is there any particular advantage of `upcomingTickets.forEach` over `for (obj of upcomingTickets)`?

src/tasks/FilterFeedTask.ts
SPGoding: Sadly no :/

src/tasks/FilterFeedTask.ts
SPGoding: No, just my personal preference here. Will changed to `forof`.

src/tasks/FilterFeedTask.ts
SPGoding: IMO the process of getting a `Channel` from the `channel` config should be handled in the task instead of in the `MojiraBot` so that the main class wouldn't be a mess. Anyway I changed it to only parse channel. The changes can be rewinded at any time.

SPGoding:
Rebased to the latest commit in `master` branch. Hope I didn't do something wrong 🤔

PR #6, opened 2020-01-11
Fix #5

### LINKED ISSUE #5: Mentioning the same issue multiple times causes a "show all tickets" link
When mentioning a ticket multiple times in one message, the bot creates a message for multiple tickets, listing only one ticket, providing a "show all n tickets" link, where n is the number of times the ticket was mentioned.

In case only the duplicated ticket was mentioned, it also gets rid of the extra information that is usually provided for a single ticket.

SPGoding:
The PR changes `ticketMatches` to a `Set`, so that tickets can't be duplicated.

------
imo using a proper data structure is more understandable than adding a statement to check whether the array already includes the said ticket.

### DIFF
--- a/src/commands/MentionCommand.ts
+++ b/src/commands/MentionCommand.ts
@@ -8,13 +8,13 @@ export default class MentionCommand extends Command {
 		const ticketRegex = RegExp( `(?:^|[^!])((?:${ BotConfig.projects.join( '|' ) })-\\d+)`, 'g' );
 
 		let ticketMatch: RegExpExecArray;
-		const ticketMatches: string[] = [];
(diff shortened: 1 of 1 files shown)


Pokechu22:
I don't have experience with TypeScript, but does `Set` maintain insertion order?  Or does it have an arbitrary order like Java's `HashSet`, where you need to use `LinkedHashSet` if you need to maintain insertion order?

SPGoding:
image

It keeps the order.

> Append value as the last element of entries.
> -- [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-set-objects)

PR #10, opened 2020-02-05
Use github actions to test build on each PR

urielsalis:


urielsalis:
This doesnt actually deploy. Its just a check.

PR #11, opened 2020-02-09
Improvements to mention ticket command

NeunEinser:
This pr includes some previously discussed changes to the mention command.

- It is now configurable whether links should trigger the bot.
- In case links trigger the bot, and a message containing only an issue link gets sent, this message is auto-deleted like usually
- It is now configurable if and which prefix is required to trigger the mention command.
- It is now configurable if and which prefixes prevent the mention command from triggering.

### DIFF
--- a/src/BotConfig.ts
+++ b/src/BotConfig.ts
@@ -27,7 +27,7 @@ export default class BotConfig {
 	public static rolesMessage: string;
 	public static requestChannels: string[];
 
-	//settings for mention command
+	// settings for mention command
 	public static ticketUrlsCauseEmbed: boolean;
 	public static requiredTicketPrefix: string;
 	public static forbiddenTicketPrefix: string;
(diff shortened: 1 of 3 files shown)


src/BotConfig.ts
```diff
@@ -68,8 +68,7 @@ export default class BotConfig {
 		if ( !settings.request_channels ) throw 'Request channels are not set';
 		this.requestChannels = settings.request_channels;
 
-		//no check, beacause it's a boolean.
-		this.ticketUrlsCauseEmbed = settings.ticketUrlsCauseEmbed;
```
NeunEinser: Was this done by mistake or does the double negation do some fucky stuff in js / ts?

src/BotConfig.ts
```diff
@@ -68,8 +68,7 @@ export default class BotConfig {
...
+		this.ticketUrlsCauseEmbed = !!settings.ticketUrlsCauseEmbed;
```
NeunEinser: Was this done by mistake or does the double negation do some fucky stuff in js / ts?

SPGoding: Convert to boolean

NeunEinser: Shouldn't TypeScript be smart enough for that? `this.ticketUrlsCauseEmbed` is declared as boolean.

violine1101: Just making sure that `this.ticketUrlsCauseEmbed` is a boolean and not something like a string or `undefined`. That wouldn't really have been too much of an effect though, at least in the way it's currently used.

The rest of the config fields is not validated either though, might do that while transitioning everything towards server-specific configs.

violine1101: And we're loading the data directly from JSON here, so `settings.ticketUrlsCauseEmbed` is of type `any`. Probably wouldn't have caused too much trouble if I've left it though. Just making sure that the bot doesn't crash if the value is undefined or otherwise faulty.

PR #17, opened 2020-02-20
Fix #9

### LINKED ISSUE #9: TypeError: Cannot read property 'hasPermission' of null when checking ModeratorPermission
Just leaving this here to save the stack trace.
```
(node:8869) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'hasPermission' of null
    at ModeratorPermission.checkPermission (.../mojira-discord-bot/src/permissions/ModeratorPermission.ts:10:17)
    at PollCommand.checkPermission (.../mojira-discord-bot/src/commands/Command.ts:18:31)
    at Function.<anonymous> (.../mojira-discord-bot/src/commands/CommandExecutor.ts:10:17)
    at Generator.next (<anonymous>)
    at .../mojira-discord-bot/src/commands/CommandExecutor.ts:8:71
    at new Promise (<anonymous>)
    at __awaiter (.../mojira-discord-bot/src/commands/CommandExecutor.ts:4:12)
    at Function.checkCommands (.../mojira-discord-bot/src/commands/CommandExecutor.ts:16:16)
    at Client.MessageEventHandler.onEvent (.../mojira-discord-bot/src/events/MessageEventHandler.ts:27:19)
    at emitOne (events.js:116:13)
UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
[DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

violine1101:
This tries to mitigate the reason for #9, in theory it seems like it happens whenever a member leaves the server before the command is processed.

I'm not really sure if that would be the only case in which this happens though, as it seems quite unlikely that this triggered the error from what I can tell. Perhaps there's something I missed?

Anyway, with this change, the permission check for moderator and owner should now always fail (and not throw an error) in case `member` is `undefined` or `null`.

### DIFF
--- a/src/permissions/ModeratorPermission.ts
+++ b/src/permissions/ModeratorPermission.ts
@@ -6,7 +6,7 @@ import Permission from './Permission';
  * This allows the command to be run by any guild member who has the "Manage messages" permission serverwide.
  */
 export default class ModeratorPermission extends Permission {
-	public checkPermission( member: GuildMember ): boolean {
-		return member.hasPermission( 'MANAGE_MESSAGES' );
+	public checkPermission( member?: GuildMember ): boolean {
+		return member?.hasPermission( 'MANAGE_MESSAGES' );
 	}
 }
--- a/src/permissions/OwnerPermission.ts
+++ b/src/permissions/OwnerPermission.ts
@@ -3,7 +3,7 @@ import Permission from './Permission';
 import BotConfig from '../BotConfig';
 
 export default class OwnerPermission extends Permission {
-	public checkPermission( member: GuildMember ): boolean {
-		return member.id === BotConfig.owner;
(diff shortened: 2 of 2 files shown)


violine1101:
As far as I'm aware, `ts-node` relies on the currently installed `typescript` package, which is required as `^3.7.2` by the `package.json` file, so it should be fine.

SPGoding:
I think enabling `strictNullChecks` in `tsconfig.json` would decrease issues like this? By turning on that option, `null` and `undefined` will be treated as separated types like this:

image

PR #19, opened 2020-02-21
Change behaviors of request channels

### LINKED ISSUE #16: Change request channel action from pinning to posting in separate channel
Currently, all requests in request channels are pinned to that channel (see #8). However, as there is a limit of 50 pinned messages on Discord (which is also why we're currently not using this feature), it might be a good idea to change it or at least provide an option for the requests to also be mirrored in a different channel, maybe if a request is not acted upon after some time passed.

SPGoding:
Resolve #16.

### DIFF
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -9,6 +9,7 @@ import MessageEventHandler from './events/MessageEventHandler';
 import AddReactionEventHandler from './events/AddReactionEventHandler';
 import RemoveReactionEventHandler from './events/RemoveReactionEventHandler';
 import MessageDeleteEventHandler from './events/MessageDeleteEventHandler';
+import MessageUpdateEventHandler from './events/MessageUpdateEventHandler';
 
 /**
  * Core class of MojiraBot
@@ -63,6 +64,7 @@ export default class MojiraBot {
 					}
 					EventRegistry.add( new AddReactionEventHandler( this.client.user.id ) );
 					EventRegistry.add( new MessageEventHandler( this.client.user.id, internalChannels ) );
+					EventRegistry.add( new MessageUpdateEventHandler( this.client.user.id, internalChannels ) );
 					EventRegistry.add( new MessageDeleteEventHandler( this.client.user.id, internalChannels ) );
 				} catch ( err ) {
 					this.logger.error( err );
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -8,6 +8,7 @@ import ErrorEventHandler from './events/ErrorEventHandler';
 import MessageEventHandler from './events/MessageEventHandler';
 import AddReactionEventHandler from './events/AddReactionEventHandler';
 import RemoveReactionEventHandler from './events/RemoveReactionEventHandler';
+import MessageDeleteEventHandler from './events/MessageDeleteEventHandler';
 
 /**
  * Core class of MojiraBot
@@ -62,6 +63,7 @@ export default class MojiraBot {
 					}
 					EventRegistry.add( new AddReactionEventHandler( this.client.user.id ) );
 					EventRegistry.add( new MessageEventHandler( this.client.user.id, internalChannels ) );
+					EventRegistry.add( new MessageDeleteEventHandler( this.client.user.id, internalChannels ) );
 				} catch ( err ) {
 					this.logger.error( err );
 				}
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -49,27 +49,31 @@ export default class MojiraBot {
 			if ( rolesChannel && rolesChannel instanceof TextChannel ) {
 				try {
 					await rolesChannel.fetchMessage( BotConfig.rolesMessage );
-
-					const internalChannels = new Map<string, TextChannel>();
-					if ( BotConfig.request.channels ) {
-						for ( let i = 0; i < BotConfig.request.channels.length; i++ ) {
-							const channelId = BotConfig.request.channels[i];
-							const internalChannelId = BotConfig.request.internal_channels[i];
-							const internalChannel = this.client.channels.get( internalChannelId );
-							if ( internalChannel && internalChannel instanceof TextChannel ) {
-								internalChannels.set( channelId, internalChannel );
-								await internalChannel.fetchMessages();
-							}
-						}
-					}
-					EventRegistry.add( new AddReactionEventHandler( this.client.user.id ) );
-					EventRegistry.add( new MessageEventHandler( this.client.user.id, internalChannels ) );
-					EventRegistry.add( new MessageUpdateEventHandler( this.client.user.id, internalChannels ) );
-					EventRegistry.add( new MessageDeleteEventHandler( this.client.user.id, internalChannels ) );
 				} catch ( err ) {
 					this.logger.error( err );
 				}
 			}
+
+			const internalChannels = new Map<string, TextChannel>();
+			if ( BotConfig.request.channels ) {
+				for ( let i = 0; i < BotConfig.request.channels.length; i++ ) {
+					const channelId = BotConfig.request.channels[i];
+					const internalChannelId = BotConfig.request.internal_channels[i];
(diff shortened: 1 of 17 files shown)


violine1101:
I don't feel like this is quite ready yet for a code review, see the discussion about the requirements of this feature in #16. I've checked out the current state of this PR and left a comment with a few thoughts on your current implementation in that issue.

There's an issue that I noticed already though: The bot will only prepare reactions for the first internal request message it ever posts, it will not prepare the reactions for any following messages to the internal channel.

SPGoding:
All known issues listed in [here](https://github.com/mojira/mojira-discord-bot/issues/16#issuecomment-590097894) have been fixed.

NeunEinser:
Some stuff I want to drop here before it gets lost in the depths of Discord:

When reacting to an internal message with multiple reactions, only the one you clicked last will be counted.
Imo, this is not how it should behave. I think it would be best to assume this was done by mistake and treat it as if the message isn't resolved.
Another solution could be to transfer all reactions.


I think it would be good to include the message text and the person who made the request in the request log.
An idea for formatting could be to have the description contain the message, linebreak, reaction (resembling a message with a reaction on Discord), add fields for creator, channel and link to request.

I don't see why channel ID and Message ID are fields in the internal request message. Since 3 fields can appear next to each other, how about splitting Go To in message link and channel, and adding one for the person making the request?

src/MojiraBot.ts
```diff
@@ -41,20 +43,29 @@ export default class MojiraBot {
...
 			const rolesChannel = this.client.channels.get( BotConfig.rolesChannel );
 			if ( rolesChannel && rolesChannel instanceof TextChannel ) {
 				try {
 					await rolesChannel.fetchMessage( BotConfig.rolesMessage );
-					for ( const channel of BotConfig.requestChannels ) {
-						const requestChannel = this.client.channels.get( channel );
-						if ( requestChannel && requestChannel instanceof TextChannel ) {
-							await requestChannel.fetchPinnedMessages();
+
+					const internalChannels = new Map<string, TextChannel>();
```
violine1101: All of this only happens if there's a roles channel. We should probably move it outside of the if block, but since I'm reworking the config stuff anyway, it's not that important (we have a roles channel set after all)

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -1,22 +1,77 @@
...
+
+	constructor( internalChannels: Map<string, TextChannel> ) {
+		this.internalChannels = internalChannels;
+	}
+
 	// This syntax is used to ensure that `this` refers to the `NewRequestEventHandler` object
-	public onEvent = ( message: Message ): void => {
-		this.logger.info( `User ${ message.author.tag } posted a new request to requests channel ${ message.channel.id }` );
+	public onEvent = async ( origin: Message ): Promise<void> => {
+		if ( origin.type !== 'DEFAULT' ) {
```
violine1101: Is it still necessary to delete system messages here? After all, the bot no longer pins messages, does it?

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -1,22 +1,77 @@
...
-				message.delete();
+		const regex = new RegExp( `(?:${MentionCommand.ticketLinkRegex.source}|${MentionCommand.ticketIdRegex.source})`, 'g' );
+
+		if ( BotConfig.request.no_link_emoji && !origin.content.match( regex ) ) {
+			origin.react( BotConfig.request.no_link_emoji );
+			const embed = new RichEmbed()
+				.setColor( '#F7C6C9' )
+				.setAuthor( origin.author.tag, origin.author.avatarURL )
+				.setDescription( `Your [request](${ origin.url }) doesn't not contain any valid ticket reference.` )
+				.setTimestamp( new Date() );
```
violine1101: Is it necessary to send an embed here? After all, the person posting the request should always immediately see it, so a ticket is not needed. Perhaps A ping could be included as well.

Maybe the bot could also mention that the user can also edit their message and does not need to delete it, along the lines of "If you want to add it, you can edit your message".

Also, is there any particular reason for using that precise color for the embed?

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -1,22 +1,77 @@
...
+		}
+
+		this.logger.info( `User ${ origin.author.tag } posted a new request to requests channel ${ origin.channel.id }` );
+
+		await origin.clearReactions();
 
-		if ( message.type === 'PINS_ADD' ) {
-			if ( message.deletable ) {
-				message.delete();
+		const regex = new RegExp( `(?:${MentionCommand.ticketLinkRegex.source}|${MentionCommand.ticketIdRegex.source})`, 'g' );
```
violine1101: Due to the possible restrictions for ticket mentions, the may think you didn't mention any ticket if you escaped the ID or didn't put the required prefix in front of it. In this case, this is not desirable in my opinion. If we decide to require a `!` in front of a ticket ID in order to mention a bot, the bot would not accept requests such as `Please resolve MC-12345 as duplicate of MC-1234`.

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -1,22 +1,77 @@
...
+			const copy = await internalChannel.send( response, embed ) as Message;
+			if ( BotConfig.request.suggested_emoji ) {
+				ReactionsUtil.reactToMessage( copy, [...BotConfig.request.suggested_emoji] );
 			}
-		} else if ( message.pinnable ) {
-			message.pin();
 		}
 	};
+
+	private replaceTicketReferenesWithRichLinks( content: string, regex: RegExp ): string {
```
violine1101: Typo: "Referenes" instead of "References"

src/events/requests/NewRequestEventHandler.ts
SPGoding: Though it could react the waiting emoji to system messages. Might change the behavior to simply ignoring them instead of deleting them?

src/events/requests/NewRequestEventHandler.ts
SPGoding: > is there any particular reason for using that precise color for the embed
Nope. I just didn't know which color to use, so I basically picked it from my avatar

src/events/requests/NewRequestEventHandler.ts
violine1101: Yeah, I think the system messages should just be ignored altogether, after all we don't do anything with them anymore, do we?

NeunEinser:
> The bot requires these fields in the embed in order to find the original message, see [`RequestsUtil`](https://github.com/mojira/mojira-discord-bot/blob/4eb86c9b165ae0c2f7f9dc3de3a1a002ff6ec878/src/util/RequestsUtil.ts).

Ah, that makes sense then.
I guess we could still think about that when the guild config is done and we might have a way of persistent storage.
I guess we could still just look at the discord url as that's also `.../server-id/channel-id/message-id`

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -33,12 +33,7 @@ export default class NewRequestEventHandler implements EventHandler {
 
 		if ( BotConfig.request.no_link_emoji && !origin.content.match( regex ) ) {
 			origin.react( BotConfig.request.no_link_emoji );
-			const embed = new RichEmbed()
-				.setColor( '#F7C6C9' )
-				.setAuthor( origin.author.tag, origin.author.avatarURL )
-				.setDescription( `Your [request](${ origin.url }) doesn't not contain any valid ticket reference.` )
-				.setTimestamp( new Date() );
-			const warning = await origin.channel.send( embed ) as Message;
+			const warning = await origin.channel.send( `${ origin.author }, your request doesn't not contain any valid ticket reference. If you'd like to add it you can edit your message.` ) as Message;
```
NeunEinser: Double negative "doesn't not" in here

SPGoding: Oops

Issue #34, opened 2020-03-12
Save logs to files


violine1101:
Generate a file for the logs every time the bot is started instead of only posting them to console, in order to be able to retrieve the logs after a crash. This should be configurable in `settings.json` for debug purposes.

PR #35, opened 2020-03-12
Added option to save log files to a directory

NeunEinser:
Implements #34 

Adds a config option for logging to a file.

* The directory can be configured in settings.json
* The file will be named in the scheme of `YYYY-MM-DDThh_mm_ss_millisZ.log`, e.g. `2020-03-12T11_24_07_488Z.log`

### DIFF
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -24,6 +24,25 @@ export default class MojiraBot {
 	private static running = false;
 
 	public static async start(): Promise<void> {
+		const logConfig: log4js.Configuration = {
+			appenders: {
+				out: { type: 'stdout' },
+			},
+			categories: {
+				default: { appenders: [ 'out' ], level: BotConfig.debug ? 'debug' : 'warn' },
+			},
+		};
+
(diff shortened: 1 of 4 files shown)


src/MojiraBot.ts
```diff
@@ -24,6 +24,25 @@ export default class MojiraBot {
 	private static running = false;
 
 	public static async start(): Promise<void> {
+		const logConfig: log4js.Configuration = {
+			appenders: {
+				out: { type: 'stdout' },
+			},
+			categories: {
+				default: { appenders: [ 'out' ], level: BotConfig.debug ? 'debug' : 'warn' },
```
NeunEinser: I was unsure here whether the default level should be `info` or `warn`.
The only thing the `debug` flag seemed to do previously was printing a log message.

violine1101: Yeah, debug mode actually doesn't really do anything currently. This way it seems good though. `warn` is a good default level for non-debug mode in my opinion, as messages with default level might or might not be important (since level is not specified), so you should definitely be able to see them.

PR #36, opened 2020-03-12
Update Github CI workflow

violine1101:


package.json
```diff
@@ -2,16 +2,21 @@
   "name": "mojira-discord-bot",
   "version": "0.0.1",
   "description": "A work-in-progress bot used to link to Mojira bug reports directly from Discord.",
-  "main": "index.js",
+  "license": "GPL-3.0",
+  "author": "violine1101",
+  "contributors": [
+    "NeunEinser",
```
urielsalis: What do we define as contributor? 😅 
Might be nice to leave in the readme so people dont just add themselves

NeunEinser: I dunno, I also dunno if you need to define that, as it's visible in git's history anyways.

violine1101: Yeah, honestly it's not too important to have this field there, it's just not like I'm the sole author of this whole thing. I've added @NeunEinser and @SPGoding since they've contributed a lot in the past few months.

We can still worry about people adding themselves here when they do so... nothing reaches the master branch until it's been approved.

PR #39, opened 2020-03-12
Allow reopening requests

### LINKED ISSUE #26: Allow for unresolving via the log
In case you accidentally resolved something and notice it too late, after the delay is already over, it would be good to have an option to un-resolve a request, for example by reacting with a certain emoji to a log message.
Doing so would then replace the emoji on the request again with the hourglass, post a log message with "un-resolved request" and send a new message into the internal requests log.

urielsalis:
Fixes #26 

Reacting with any emoji on a request log in the log channel will reopen the request

(Newbie in typescript so please review carefully)

### DIFF
--- a/src/events/AddReactionEventHandler.ts
+++ b/src/events/AddReactionEventHandler.ts
@@ -1,9 +1,9 @@
-import {User, MessageReaction, TextChannel} from 'discord.js';
+import { User, MessageReaction, TextChannel } from 'discord.js';
 import BotConfig from '../BotConfig';
 import EventHandler from './EventHandler';
 import SelectRoleEventHandler from './roles/SelectRoleEventHandler';
-import ResolveRequstEventHandler from './requests/ResolveRequestEventHandler';
-import NewRequestEventHandler from "./requests/NewRequestEventHandler";
+import ResolveRequestEventHandler from './requests/ResolveRequestEventHandler';
+import NewRequestEventHandler from './requests/NewRequestEventHandler';
 
 export default class AddReactionEventHandler implements EventHandler {
 	public readonly eventName = 'messageReactionAdd';
@@ -11,7 +11,7 @@ export default class AddReactionEventHandler implements EventHandler {
 	private readonly botUserId: string;
 
 	private readonly selectRoleHandler = new SelectRoleEventHandler();
-	private readonly resolveRequestHandler = new ResolveRequstEventHandler();
+	private readonly resolveRequestHandler = new ResolveRequestEventHandler();
 	private readonly newRequestEventHandler: NewRequestEventHandler;
 
 	constructor( botUserId: string, internalChannels: Map<string, TextChannel> ) {
--- a/src/events/AddReactionEventHandler.ts
+++ b/src/events/AddReactionEventHandler.ts
@@ -32,7 +32,7 @@ export default class AddReactionEventHandler implements EventHandler {
 			this.resolveRequestHandler.onEvent( messageReaction, user );
 		} else if ( BotConfig.request.log_channel.includes( messageReaction.message.channel.id ) ) {
 			// Handle reopening a user request
-			this.resolveRequestHandler.onReopen( messageReaction, user, this.newRequestEventHandler );
+			this.newRequestEventHandler.onReopen( messageReaction, user );
 		}
 	};
 }
--- a/src/events/AddReactionEventHandler.ts
+++ b/src/events/AddReactionEventHandler.ts
@@ -1,8 +1,9 @@
-import { User, MessageReaction } from 'discord.js';
+import {User, MessageReaction, TextChannel} from 'discord.js';
 import BotConfig from '../BotConfig';
 import EventHandler from './EventHandler';
 import SelectRoleEventHandler from './roles/SelectRoleEventHandler';
 import ResolveRequstEventHandler from './requests/ResolveRequestEventHandler';
+import NewRequestEventHandler from "./requests/NewRequestEventHandler";
 
 export default class AddReactionEventHandler implements EventHandler {
 	public readonly eventName = 'messageReactionAdd';
@@ -11,9 +12,11 @@ export default class AddReactionEventHandler implements EventHandler {
 
 	private readonly selectRoleHandler = new SelectRoleEventHandler();
 	private readonly resolveRequestHandler = new ResolveRequstEventHandler();
+	private readonly newRequestEventHandler: NewRequestEventHandler;
 
-	constructor( botUserId: string ) {
+	constructor( botUserId: string, internalChannels: Map<string, TextChannel> ) {
 		this.botUserId = botUserId;
+		this.newRequestEventHandler = new NewRequestEventHandler( internalChannels );
 	}
 
 	// This syntax is used to ensure that `this` refers to the `AddReactionEventHandler` object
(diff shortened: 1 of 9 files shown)


src/events/requests/NewRequestEventHandler.ts
```diff
@@ -28,6 +31,53 @@ export default class NewRequestEventHandler implements EventHandler {
...
+		const logMessage = reaction.message;
+		const embeds = logMessage.embeds;
+		if( embeds.length == 0 ) {
+			const warning = await logMessage.channel.send( `${ logMessage.author }, this is not a valid log message.` ) as Message;
+			warning.delete( BotConfig.request.no_link_warning_lifetime || 0 );
+		}
+
+		// Assume first embed is the log message
+		const logEmbed = embeds[0];
+		const url: string = logEmbed.fields[logEmbed.fields.length - 1].value;
```
SPGoding: If we managed to parse the message links, I guess we can remove those two fields (`Channel` and `Message`) in internal request messages and rewrite the urils [here](https://github.com/mojira/mojira-discord-bot/blob/89568e7482f58e2974b7fc2443c042cf608e929c/src/util/RequestsUtil.ts#L5).

src/events/AddReactionEventHandler.ts
```diff
@@ -1,8 +1,9 @@
-import { User, MessageReaction } from 'discord.js';
+import {User, MessageReaction, TextChannel} from 'discord.js';
```
NeunEinser: The code style requires spaces here.
You can install the vscode extention `ESLint` to get those marked automatically.
Also you can run `npm run lint` to fix those things.

src/events/AddReactionEventHandler.ts
```diff
@@ -1,8 +1,9 @@
...
 import BotConfig from '../BotConfig';
 import EventHandler from './EventHandler';
 import SelectRoleEventHandler from './roles/SelectRoleEventHandler';
 import ResolveRequstEventHandler from './requests/ResolveRequestEventHandler';
+import NewRequestEventHandler from "./requests/NewRequestEventHandler";
```
NeunEinser: Always use single quotes, because that's convention for some reason (also caught by `lint`)

src/events/requests/NewRequestEventHandler.ts
urielsalis: Can you elaborate? not sure I got that 😅

src/events/AddReactionEventHandler.ts
```diff
@@ -11,9 +12,11 @@ export default class AddReactionEventHandler implements EventHandler {
 
 	private readonly selectRoleHandler = new SelectRoleEventHandler();
 	private readonly resolveRequestHandler = new ResolveRequstEventHandler();
+	private readonly newRequestEventHandler: NewRequestEventHandler;
 
-	constructor( botUserId: string ) {
+	constructor( botUserId: string, internalChannels: Map<string, TextChannel> ) {
```
NeunEinser: This shouldn't be needed as it's statically available in `BotConfig.request.internal_channels`

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -1,10 +1,11 @@
-import { Message, TextChannel, RichEmbed } from 'discord.js';
+import {Message, TextChannel, RichEmbed, MessageReaction, User} from 'discord.js';
```
NeunEinser: Spaces

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -28,6 +29,53 @@ export default class NewRequestEventHandler implements EventHandler {
 
 		this.logger.info( `User ${ origin.author.tag } posted a new request to requests channel ${ origin.channel.id }` );
 
+		this.handleNewRequest( origin );
+	};
+
+	// This syntax is used to ensure that `this` refers to the `NewRequestEventHandler` object
+	public onReopen = async ( reaction: MessageReaction, user: User ): Promise<void> => {
+		this.logger.info( `User ${user.tag} is reopening the request message '${reaction.message.id}'` );
```
NeunEinser: That's some spaces for `${user.tag}` and the other one, `lint` won't mark you, but they are generally there in other parts of the code.
I don't think they are necessary but I am forcing myself to add those now as well, to keep it consistent
Not gonna mark them in other instances tho.

src/events/requests/NewRequestEventHandler.ts
NeunEinser: Agreed. Not needed for this PR tho imo

src/events/requests/NewRequestEventHandler.ts
```diff
@@ -28,6 +29,53 @@ export default class NewRequestEventHandler implements EventHandler {
...
+						const log = new RichEmbed()
+							.setColor( '#F7C6C9' )
+							.setAuthor( requestMessage.author.tag, requestMessage.author.avatarURL )
+							.setDescription( requestMessage.content )
+							.addField( 'Channel', requestMessage.channel.toString(), true )
+							.addField( 'Message', `[Here](${requestMessage.url})`, true )
+							.setFooter( `${user.tag} reopened this request`, user.avatarURL )
+							.setTimestamp( new Date() );
+						logChannel.send( log );
+					}
```
NeunEinser: I feel like this needs to be distinguished more from the normal kind of log message.
Probably should at least have a different color.

src/events/requests/NewRequestEventHandler.ts
NeunEinser: @urielsalis Currently Channel and Message are fields that are present on embeds from requests, which are used for the bot to keep track of the original message.
You are simply getting those IDs from the url, which could be done in the requests channel too to replace those fields with something more useful.

At least that's how I understood it ;)

src/events/AddReactionEventHandler.ts
urielsalis: Weird, I ran npm run lint before committing

Also I use intellij idea which marks them as errors supposedely

src/events/AddReactionEventHandler.ts
urielsalis: Tried replacing it, apparently not the same types. This is also how all the otehr listeners are doing it

src/events/requests/NewRequestEventHandler.ts
urielsalis: Any suggestion for the color?

src/events/AddReactionEventHandler.ts
violine1101: Yeah, lint doesn't catch missing spaces around curly braces. it only works with parentheses. There's probably some rule for that somewhere that we could enable, though.

src/events/AddReactionEventHandler.ts
NeunEinser: Ah actually nevermind, the map is not stored and only created in `MojiraBot` as it lokks loke

src/events/requests/NewRequestEventHandler.ts
NeunEinser: I dunno, since the other one is a weird red, maybe just green? ¯\_(ツ)_/¯

src/events/requests/NewRequestEventHandler.ts
urielsalis: ah, so you mean adding a channel id and message id fields in the log channel and reading from there instead of url?

SPGoding: Ah sorry I didn't say it clearly. Basically what I meant was the thing NeunEinser said, we could remove the two ugly IDs in the embeds of pending requests. That doesn't have to be done in this PR though 
No changes shall be needed in your code, it is just about simplifying the former implementation of request channels made by me ;D

src/events/requests/NewRequestEventHandler.ts
urielsalis: Changed to the triadic green of the current color

urielsalis:
Can this be merged @violine1101 ?

src/events/requests/NewRequestEventHandler.ts
violine1101: Yeah, generally I think it's good to have spaces in these places, but it ESLint doesn't have the rule about this enabled yet. I think I'll configure the rule and change the existing code when merging so that it'll have spaces everywhere, just like with the brackets.

PR #48, opened 2020-04-26
The variable {{message}} of response messages is now replaced with a quoted message

### LINKED ISSUE #47: Request block comments don't work correctly
When a request has multiple lines, the reply doesn't properly adjust to that. For instance if a request looks like:
```
MC-1234 duplicates
- MC-5678
- MC-9012
- MC-3456
```
The resulting answer template would be
```
> MC-1234 duplicates
- MC-5678
- MC-9012
- MC-3456

<answer>
```
instead of
```
> MC-1234 duplicates
> - MC-5678
> - MC-9012
> - MC-3456

<answer>
```

NeunEinser:
Fixes #47.
Requires config change, as `{{message}}` in `response_message` will now be replaced with a properly quoted message instead of just the message.

### DIFF
--- a/settings-template.jsonc
+++ b/settings-template.jsonc
@@ -95,10 +95,10 @@
 		// Parameters are {{autor}}, {{url}} and {{message}}.
 		// {{author}} will be replaced by "@DiscordUser#1234"
 		// {{url}} will be replaced by  the link to the message https://discordapp.com/channels/.../.../...
-		// {{message}} will be replaced by a qoute of the contents of the message.
+		// {{message}} will be replaced by a quote of the contents of the message.
 		//
 		// Optional. Defaults to an empty message.
-		"response_message": "```\n{{author}} {{url}}\n{{message}}\n \n```"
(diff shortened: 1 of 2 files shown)


settings-template.jsonc
```diff
@@ -95,10 +95,10 @@
 		// Parameters are {{autor}}, {{url}} and {{message}}.
 		// {{author}} will be replaced by "@DiscordUser#1234"
 		// {{url}} will be replaced by  the link to the message https://discordapp.com/channels/.../.../...
-		// {{message}} will be replaced by a qoute of the contents of the message.
+		// {{message}} will be replaced by a quote of the contents of the message.
 		//
 		// Optional. Defaults to an empty message.
-		"response_message": "```\n{{author}} {{url}}\n{{message}}\n \n```"
+		"response_message": "```\n{{author}} {{url}}\n{{message}}\n\n```"
```
NeunEinser, line 101: I think that was intentionally, because without the space the block doesn't actually have the new line, as the trailing newline of the code block gets stripped

PR #65, opened 2020-06-28
Forward requests posted during downtime

### LINKED ISSUE #24: The bot currently does not seem to post old requests to the internal request channel
This could potentially become an issue when requests are posted during a downtime.
It would be good if the bot would scan for messages without any reaction in the request channel after starting.

SPGoding:
Fix #24. Tested on my local server.


SPGoding:
Oops, I mapped the wrong variable when merging.

Issue #49, opened 2020-04-29
Automatically announce when an unreleased version is renamed or marked as released


violine1101:
Additionally to announcing whenever a ticket has been fixed, it would also be nice if the bot was able to announce when an unreleased version has been released or renamed, because that usually signifies that that version is about to release.

PR #66, opened 2020-07-02
Add version feed (#49)

violine1101:
This PR adds a version feed which announces when a version is renamed, or if its properties change (i.e. it gets released or archived).

Here's a screenshot from my testing:

grafik

### DIFF
--- a/settings-template.jsonc
+++ b/settings-template.jsonc
@@ -140,5 +140,20 @@
 			"title_single": "One ticket blabla in the last 15 minutes!"
 		}
 		// ... you can add more feeds here
+	],
+
+	"version_feed_interval": 900000,
+
+	"version_feeds": [
+		{
+			// The project whose versions should be monitored
+			"project": "<Project key>",
+
+			// The channel in which the feed should be sent
+			"channel": "<Feed channel ID>",
+
+			// How many versions should be monitored; only the x latest versions are monitored
+			"scope": 5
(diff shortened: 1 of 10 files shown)


violine1101:
I have read the CLA Document and I hereby sign the CLA

settings-template.jsonc
```diff
@@ -140,5 +140,20 @@
...
 		}
 		// ... you can add more feeds here
+	],
+
+	"version_feed_interval": 900000,
+
+	"version_feeds": [
+		{
+			// The project whose versions should be monitored
+			"project": "<Project key>",
```
NeunEinser, line 150: I wonder if it might make more sense to just monitor all projects in one feed.
Releses don't happen that often, and I think in #news it works pretty well.

violine1101: I thought it would fit best in the `#java-fixes` channel. I'm not sure how other projects do versions. For example on MCD, versions often only get added after the version has been released. Not as exciting.

But apart from that, this is a technical limitation, JIRA's API only allows for monitoring the versions for only one specific project at once since versions are bound to a project. But if we want to have a feed for version changes in multiple (or all) projects, we could add multiple feeds (even if they all point to the same channel).

violine1101:
FYI, I've already set the new required settings server-side, so this can be merged without the bot going down.

PR #87, opened 2020-08-06
Make version feed actions configurable

violine1101:
This makes version feeds completely configurable, so that we can decide what actions we want to have included in the feed.

### DIFF
--- a/config/main.yml
+++ b/config/main.yml
@@ -52,3 +52,6 @@ versionFeeds:
   - project: MC
     channel: '666349583227682819'
     scope: 5
+    actions:
+      - released
+      - renamed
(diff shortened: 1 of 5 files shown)


config/main.yml
```diff
@@ -51,4 +51,7 @@ filterFeeds:
 versionFeeds:
   - project: MC
     channel: '666349583227682819'
-    scope: 5
\ No newline at end of file
+    scope: 5
+    actions:
+      - released
+      - renamed
```
NeunEinser, line 57: Might be worth adding "unreleased" too (as, if that ever happens, we would end up sending the release message twice, making it look like a bug)

violine1101: Hmm, yeah, that has never happened before, but probably worth adding anyway in case something goes wrong.

PR #89, opened 2020-08-08
Mention users that use !jira ping in the response message

### LINKED ISSUE #72: Ping user of '!jira ping' command when responding with 'Pong!'
Occasionally, a user deletes their !jira ping comment, leaving a random 'Pong!' by MojiraBot in #bot-commands. It would be better if the bot could ping the user of the command (for example "@markderickson Pong!"). This will make it easier to see that the pong wasn't just a reaction to an older message, but a reaction to a deleted message.

An example of this can be found at https://discordapp.com/channels/647810384031645728/647814935572643840/739336733002301542.

chandler05:
This should now be reading "@user Pong!"

Fixes #72

### DIFF
--- a/src/commands/PingCommand.ts
+++ b/src/commands/PingCommand.ts
@@ -10,7 +10,7 @@ export default class PingCommand extends PrefixCommand {
 		}
 
 		try {
-			await message.channel.send( 'Pong!' );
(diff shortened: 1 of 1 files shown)


chandler05:
I have read the CLA Document and I hereby sign the CLA

PR #91, opened 2020-08-08
React to all feed messages with 🎉

### LINKED ISSUE #71: Automatically add :tada: reaction to any message in a Fixes channel
This doesn't really need to be a thing, and its not important, but I thought it would be kinda helpful for there to automatically be a :tada: reaction to fixes, because users add that reaction to every fix and it would be easier to be able to just click it instead of having to search for the reaction the first time

chandler05:
Fixes #71 
🎉

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -42,9 +42,6 @@ request:
     > {{message}}
     
     ```
-feed:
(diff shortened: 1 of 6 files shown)


chandler05:
I hope I did this right 🤞

PR #94, opened 2020-08-08
Fix #7 (Bot does not reconnect after being disconnected)

### LINKED ISSUE #7: Bot does not reconnect after being disconnected
After the bot is disconnected for whatever reason, it does not try to reconnect.

violine1101:
So apparently, this branch was forgotten about since January. I just came across it while cleaning up some local branches.

## Purpose
Try to reconnect if the client gets disconnected from Discord for whatever reason.

## Approach
Listen for the disconnect event, and when it happens, try to reconnect every 10 seconds for 10 minutes. If after that the bot is still unable to reconnect, shut down the bot.

## Future work
When we upgrade to discord.js 12, we need to investigate whether there's some better (perhaps built-in?) way to handle this.

### DIFF
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -13,6 +13,7 @@ import MessageDeleteEventHandler from './events/MessageDeleteEventHandler';
 import MessageUpdateEventHandler from './events/MessageUpdateEventHandler';
 import VersionFeedTask from './tasks/VersionFeedTask';
 import NewRequestEventHandler from './events/requests/NewRequestEventHandler';
+import DisconnectEventHandler from './events/DisconnectEventHandler';
 
 /**
  * Core class of MojiraBot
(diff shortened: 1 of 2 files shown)


src/MojiraBot.ts
```diff
@@ -42,7 +43,7 @@ export default class MojiraBot {
 			// Register events.
 			EventRegistry.setClient( this.client );
 			EventRegistry.add( new ErrorEventHandler() );
-			EventRegistry.add( new RemoveReactionEventHandler( this.client.user.id ) );
```
NeunEinser, line 45: Why was that event handler removed?

violine1101: Ah, good catch. Something must have gone wrong during rebasing.

violine1101:
I have come to the conclusion that this PR is no longer needed. The `disconnect` event appears to have been deprecated and is only present in the typings from what I can tell.

PR #95, opened 2020-08-08
Add a !jira moo command

chandler05:
## Purpose
Creates an embed for MC-772

### DIFF
--- a/src/commands/MooCommand.ts
+++ b/src/commands/MooCommand.ts
@@ -13,9 +13,9 @@ export default class MooCommand extends PrefixCommand {
 
 		try {
 			new SingleMention( 'MC-772' );
-			let embed: RichEmbed; 
+			let embed: RichEmbed;
 			await message.channel.send( embed );
-			await ReactionsUtil.reactToMessage( embed, ['🐄','🥛'] );
+			await ReactionsUtil.reactToMessage( embed, ['🐄' , '🥛'] );
 		} catch {
 			return false;
 		}
--- a/src/commands/MooCommand.ts
+++ b/src/commands/MooCommand.ts
@@ -15,7 +15,7 @@ export default class MooCommand extends PrefixCommand {
 			new SingleMention( 'MC-772' );
 			let embed: RichEmbed;
 			await message.channel.send( embed );
-			await ReactionsUtil.reactToMessage( embed, ['🐄' , '🥛'] );
+			await ReactionsUtil.reactToMessage( embed, ['🐄', '🥛'] );
 		} catch {
 			return false;
 		}
--- a/src/commands/MooCommand.ts
+++ b/src/commands/MooCommand.ts
@@ -15,7 +15,7 @@ export default class MooCommand extends PrefixCommand {
 			new SingleMention( 'MC-772' );
 			let embed: RichEmbed;
 			await message.channel.send( embed );
-			await ReactionsUtil.reactToMessage( embed, ['🐄', '🥛'] );
+			await ReactionsUtil.reactToMessage( message, ['🐄', '🥛'] );
 		} catch {
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


src/commands/MooCommand.ts
```diff
@@ -0,0 +1,37 @@
+import { Message } from 'discord.js';
+import PrefixCommand from './PrefixCommand';
+import { SingleMention } from '../mentions/SingleMention';
+
+export default class PingCommand extends PrefixCommand {
```
violine1101, line 5: You forgot to rename this to "MooCommand" here. Not a big deal since it's a default import though.

src/commands/MooCommand.ts
```diff
@@ -0,0 +1,37 @@
+import { Message } from 'discord.js';
+import PrefixCommand from './PrefixCommand';
+import { SingleMention } from '../mentions/SingleMention';
+
+export default class PingCommand extends PrefixCommand {
+	public readonly aliases = ['moo', 'test'];
```
violine1101, line 6: The "test" alias is already taken by !jira ping

src/commands/MooCommand.ts
```diff
@@ -0,0 +1,37 @@
...
+export default class PingCommand extends PrefixCommand {
+	public readonly aliases = ['moo', 'test'];
+
+	public async run( message: Message, args: string ): Promise<boolean> {
+		if ( args.length ) {
+			return false;
+		}
+
+		try {
+			new SingleMention("MC-772");
```
violine1101, line 14: You only created the mention object here, however you still need to get the embed from it and send the embed in a message. The mention object doesn't do anything on its own.

src/commands/MooCommand.ts
```diff
@@ -0,0 +1,37 @@
...
+		} catch {
+			return false;
+		}
+
+		try {
+			await message.react( '🐄' );
```
violine1101, line 20: Should these reactions be added to the bot's message or to the original command? Currently they're added to the original command.

Also, this code works, but you could also look into replacing this by the `reactToMessage` method from `ReactionsUtil`, which is made for adding multiple emojis at once.

src/commands/MooCommand.ts
violine1101: Also, this part of the code doesn't follow the style guide. You can use the command `npm run lint` to check if your code is formatted correctly.

src/commands/MooCommand.ts
chandler05: Renamed it to MooCommand

src/commands/MooCommand.ts
chandler05: switched it to "cow"

src/commands/MooCommand.ts
chandler05: Fixed styling/formatting

src/commands/MooCommand.ts
chandler05: Switched both of these over

src/commands/MooCommand.ts
chandler05: It now sends the message (hopefully)

src/commands/MooCommand.ts
```diff
@@ -0,0 +1,29 @@
...
+
+	public async run( message: Message, args: string ): Promise<boolean> {
+		if ( args.length ) {
+			return false;
+		}
+
+		try {
+			new SingleMention( 'MC-772' );
+			let embed: RichEmbed;
+			await message.channel.send( embed );
```
NeunEinser, line 17:
```suggestion
			const mention = new SingleMention( 'MC-772' );
			const embed = await mention.getEmbed();
```
Just fixing 😛

src/commands/MooCommand.ts
```diff
@@ -0,0 +1,29 @@
...
+	public readonly aliases = ['moo', 'cow'];
+
+	public async run( message: Message, args: string ): Promise<boolean> {
+		if ( args.length ) {
+			return false;
+		}
+
+		try {
+			new SingleMention( 'MC-772' );
+			let embed: RichEmbed;
```
NeunEinser, line 16:
```suggestion
			const mention = new SingleMention( 'MC-772' );
			const embed = await mention.getEmbed();
```
Just fixing 😛

PR #99, opened 2020-08-16
Change client settings to fix #32

### LINKED ISSUE #32: Bot does not register role removals after being restarted
When removing a reaction from a role selection message, the bot no longer seems to be able to register that removal and won't remove the role, in case the role reaction was added before the bot was started.

chandler05:
## Purpose
Fix #32
## Approach
I just did what SPGoding said in their comment on the issue, hopefully that was it


violine1101:
That alone won't work. Enabling partials requires a rewrite of the entire reaction handling code. This is not easy.

PR #106, opened 2020-09-10
Allow multiple role selection messages

SPGoding:
## Purpose
Allow multiple role selection groups. And added a new role group for pronouns.

## Future work
After merging this PR, the bot would send a message in the current `#role-selection` channel automagically. We would want to save the message ID of that message to the config's `message` field, so that the bot can find that message after restart.

This isn't ideal, but I'm just too lazy to implement the auto-saving-config feature.

### DIFF
--- a/src/util/RoleSelectionUtil.ts
+++ b/src/util/RoleSelectionUtil.ts
@@ -14,7 +14,7 @@ export class RoleSelectionUtil {
 			.setColor( 'AQUA' );
 		for ( const role of groupConfig.roles ) {
 			const roleEmoji = MojiraBot.client.emojis.cache.get( role.emoji );
-			const textEmoji = ( roleEmoji == undefined ) ? '❓' : roleEmoji.toString();
+			const textEmoji = ( roleEmoji == undefined ) ? role.emoji : roleEmoji.toString();
 			embed.addField( textEmoji, role.desc );
 		}
 		let sentMessage: Message | Message[];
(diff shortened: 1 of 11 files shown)


src/util/RoleSelectionUtil.ts
```diff
@@ -0,0 +1,40 @@
...
+				this.logger.error( 'Result of sending role selection message was not exactly one message' );
+				return;
+			} else {
+				sentMessage = sentMessage[0];
+			}
+		}
+		ReactionsUtil.reactToMessage( sentMessage as Message, groupConfig.roles.map( role => role.emoji ) );
+
+		// TODO: Ideally we would like to save the message ID automagically.
+		this.logger.warn( `Please set the 'messageId' for role selection group '${ groupConfig.prompt }' to '${ ( sentMessage as Message ).id }' in the config.` );
```
NeunEinser: Is the message ID saved to ram, so we don't need to restart the bot after updating the yaml?

SPGoding: We do need to edit the yaml and restart the bot after it sends the message, because the bot reads the config in `RoleSelectEventHandler` and `RoleRemoveEventHandler`.

SPGoding: And I just tried that there would be an "object is not extensible" error if I tried to set groupConfig.message; not sure why that happened ;/

NeunEinser (approved):

src/util/RoleSelectionUtil.ts
NeunEinser: Okay. It's not a big deal, we only have to do it once anyway, so not really worth the effort to make it pretty.

PR #111, opened 2020-09-11
Fix version log and catch promise errors properly

violine1101:
## Purpose
This fixes the bot spamming `#version-log` with supposed version changes indefinitely.

## Approach
This makes feed reactions optional, meaning that they can be omitted from the config. Previously the reactions were not optional for feeds and if they were not defined, the bot would get stuck in a loop.

Additionally, for better error reporting, all promises are now properly handled with `await` and `try`/`catch` where necessary.

## Future work
Improve the config further so that "sub-configs" like the filter config are type-safe as well.

### DIFF
--- a/src/events/request/RequestEventHandler.ts
+++ b/src/events/request/RequestEventHandler.ts
@@ -34,15 +34,30 @@ export default class RequestEventHandler implements EventHandler<'message'> {
 		const regex = new RegExp( `(?:${ MentionCommand.ticketLinkRegex.source }|(${ MentionCommand.ticketPattern }))(\\?\\S+)?`, 'g' );
 
 		if ( BotConfig.request.noLinkEmoji && !origin.content.match( regex ) ) {
-			origin.react( BotConfig.request.noLinkEmoji );
-			const timeout = BotConfig.request.noLinkWarningLifetime;
-			const warning = await origin.channel.send( `${ origin.author }, your request doesn't contain any valid ticket reference. If you'd like to add it you can edit your message.` ) as Message;
-			warning.delete( { timeout } );
+			try {
+				await origin.react( BotConfig.request.noLinkEmoji );
+			} catch ( error ) {
+				this.logger.error( error );
+			}
+
+			try {
+				const warning = await origin.channel.send( `${ origin.author }, your request doesn't contain any valid ticket reference. If you'd like to add it you can edit your message.` );
+
+				const timeout = BotConfig.request.noLinkWarningLifetime;
+				await warning.delete( { timeout } );
+			} catch ( error ) {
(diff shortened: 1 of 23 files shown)


src/events/request/RequestEventHandler.ts
```diff
@@ -34,15 +34,30 @@ export default class RequestEventHandler implements EventHandler<'message'> {
 		const regex = new RegExp( `(?:${ MentionCommand.ticketLinkRegex.source }|(${ MentionCommand.ticketPattern }))(\\?\\S+)?`, 'g' );
 
 		if ( BotConfig.request.noLinkEmoji && !origin.content.match( regex ) ) {
-			origin.react( BotConfig.request.noLinkEmoji );
-			const timeout = BotConfig.request.noLinkWarningLifetime;
-			const warning = await origin.channel.send( `${ origin.author }, your request doesn't contain any valid ticket reference. If you'd like to add it you can edit your message.` ) as Message;
-			warning.delete( { timeout } );
+			try {
```
SPGoding, line 37: Should line `L32` in this file be in a `try-catch` block?

src/events/roles/RoleSelectEventHandler.ts
```diff
@@ -17,7 +17,7 @@ export default class RoleSelectEventHandler implements EventHandler<'messageReac
 		const role = group.roles.find( searchedRole => searchedRole.emoji === messageReaction.emoji.id || searchedRole.emoji === messageReaction.emoji.name );
 
 		if ( !role ) {
-			messageReaction.users.remove( user );
+			await messageReaction.users.remove( user );
```
SPGoding, line 20: Should this be put in a `try-catch` block?

src/tasks/FilterFeedTask.ts
```diff
@@ -74,8 +74,16 @@ export default class FilterFeedTask extends Task {
 
 					if ( this.channel instanceof TextChannel ) {
 						const filterFeedMessage = await this.channel.send( message, embed );
-						NewsUtil.publishMessage( filterFeedMessage );
-						filterFeedMessage.react( this.filterFeedEmoji );
+
```
SPGoding, line 77: Should `L77` be put in a `try-catch` block?

SPGoding (approved):

src/events/request/RequestEventHandler.ts
violine1101: probably should be, yeah

src/tasks/FilterFeedTask.ts
violine1101: Doesn't need to be, `NewsUtil.publishMessage` doesn't throw

PR #122, opened 2020-11-08
Fix backtracking requests

SPGoding:
## Purpose
Fix backtracking requests. Now the bot will backtrack to the last request that was reacted by the bot. Requests that have already been reacted by human beings won't be forwarded.

Also added the process ID to the footer of filter feed.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -43,5 +43,4 @@ request:
     {{author}} {{url}}
     {{message}}
     
-    <Enter your response here>
     ```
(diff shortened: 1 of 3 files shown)


violine1101 (commented):
Two fundamental issues with this:
* The bot always checks all the messages in the request channels on startup.
* The bot first handles the messages posted last. This means that if the bot is stopped while it's still catching up, the next time it won't catch on the messages it initially missed.

Because I wanted to get rid of this infinite loop I went ahead and refactored the whole thing (PR #124). I hope you don't mind 😅

violine1101:
Superseded by #124.

PR #129, opened 2020-11-30
Add an emoji to show the prependResponseMessage without resolving the request

### LINKED ISSUE #121: Show response template without the need to resolve a pending request
## The Problem
When you don't want to resolve a pending request but just copy the response template, you currently need to react to one of the options, copy the template, and remove the reaction again.

## Possible Solution
Add an additional reaction to pending requests that is just used to show the response template.

chandler05:
Closes #121 

A similar workflow to what was mentioned in the comments on that issue, with a few changes:

1. user triggers the dedicated "reply" reaction on the internal request (:speech_balloon: emoji)
2. bot edits internal request to add the preview
3. bot does **not** remove preview until...
4. user removes the dedicated "reply" reaction
5. bot removes the preview

I have set it up so that the user does not have to worry about the request being resolved, and so that the user also does not have to act that fast to get the message. The only thing to remember is the user still has to remove the speech bubble in order to remove the message.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -31,8 +31,10 @@ request:
     - ✅
     - ☑️
     - ❌
+    - 💬
   
   ignorePrependResponseMessageEmoji: ✅
+  ignoreResolutionEmoji: 💬
 
   resolveDelay: 10000
   prependResponseMessage: whenResolved
--- a/config/default.yml
+++ b/config/default.yml
@@ -31,10 +31,10 @@ request:
     - ✅
     - ☑️
     - ❌
-    - 💬
+    - :speech_balloon:
   
   ignorePrependResponseMessageEmoji: ✅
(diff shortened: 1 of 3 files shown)


Jack-McKalling:
This suggested workflow seems more convenient than what I suggested there. But I can't see if these commits actually indeed follow it, it looks to me it's actually following my suggestion. Unless I just don't know where the code is for (responding to) removing the reaction again? I guess I have to really dive into the whole project before I can review appropriately. But thanks so far!

chandler05:
The removal of the reaction is handled at two other places, the first one being the RequestUnresolveEventHandler and the deletion of the internal request message. I have a check in place to make sure that the internal message is not removed if you have used the speech balloon. However, the response template will still be removed because I didn't change the RequestUnresolveEventHandler

SPGoding (approved):

PR #131, opened 2020-11-30
Show an empty line in response message

SPGoding:

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -44,5 +44,5 @@ request:
     ```
     {{author}} {{url}}
     {{message}}
-    .
+     
     ```
--- a/config/default.yml
+++ b/config/default.yml
@@ -44,5 +44,5 @@ request:
     ```
     {{author}} {{url}}
     {{message}}
-    
+    .
     ```


chandler05:
Why use a period to create the empty line?

chandler05 (approved):

SPGoding:
> 
> 
> Why use a period to create the empty line?

Because we need something to keep showing this line.

Pokechu22:
What about a non-breaking space character (`&nbsp;`)?

SPGoding:
Doesn't seem to work. Though I changed it to use hair space (` `) instead.

Jack-McKalling:
Is this really necessary, for instance for mobile? Because I always already get a newline. Just triple-click to select the full line (including newline feed) and drag to select both lines. On mobile you can't click, so we really need a "Copy to clipboard" button

violine1101:
Hmm, well personally I never had an issue with the line break on that message, I use the same technique as Jack to copy the message.

The issue with this fix is that it adds an additional zero width space character that you might not want to have in your message, you'd need to delete it every time.

SPGoding:
I got used to Jack's method as well, so I guess we could just close this lol.

PR #133, opened 2020-12-01
Send DMs if a user's request is resolved and depending on a role selection

chandler05:
## Future work
I need IDs for the three roles (which I can't create) and a message ID from #role-selection (same issue)

### DIFF
--- a/config/main.yml
+++ b/config/main.yml
@@ -71,6 +71,22 @@ roleGroups:
           Other Pronoun
           (please indicate in your nickname)
         emoji: '🇴'
+  - prompt: |-
+      Select a reaction on this message if you want to be notified in DMs when your requests are resolved.
+    channel: '648479533246316555'
+    message: #TODO
+    radio: true
+    roles:
+      - id: #TODO
+      desc: Be notified of all of your resolved requests.
+      emoji: '🔔'
+      - id: #TODO
+      desc: Be notified of your resolved requests that are not resolved as ✅.
+      emoji: '🛎️'
+      - id: #TODO
+      desc: Be notified of your resolved requests that are resolved at least three days after they were created.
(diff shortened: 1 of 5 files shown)


config/main.yml
```diff
@@ -71,6 +71,35 @@ roleGroups:
...
+      - id: #TODO
+      desc: Be notified of your resolved requests that are resolved as ✅.
+      emoji: '✅'
+      - id: #TODO
+      desc: Be notified of your resolved requests that are not resolved as ✅.
+      emoji: '☑️'
+  - prompt: |-
+      Request Notifications (2/2)
+      Select a reaction on this message if you want to be notified in DMs when requests are resolved after a certain amount of time.
+      Default is no time requirement.
```
pine1needle, line 90:
```suggestion
      Select a reaction on this message if you want to only be notified about requests that are at least __ days old when they are resolved.
      No reaction means there is no time requirement.
```
I think the prompt I wrote above might be more clear. The prompt should tell the reader that the unit of time is days, and I think "no reaction" is clearer than "default".

pine1needle: I just realized that the description of each role/emoji specifies the amount of time that each emoji means, so specifying the unit of time in the prompt isn't as necessary as I had originally thought.

chandler05: I added the "no reaction" message you suggested

violine1101:
I'm not sure whether this feature is really needed still. As I said previously I'd like to avoid adding extra complexity to MojiraBot at this point in time. I don't think there's reasonable demand for this feature, but I could be wrong.

Also another merge is needed which would probably have some merge conflicts as well.

I'd like to propose to postpone this feature. Please create an issue for it (seems like we don't have one already) so we can decide later if this is still something that we want or need.

PR #134, opened 2020-12-01
Remove extra reactions from the request channels

### LINKED ISSUE #58: Automatically remove emojis added to a requests section by other users
#java-requests - Google Chrome 6_26_2020 11_48_37 AM (2)
As you can see from the image, users are able to add an emoji of the same type to a request after the bot already has added one (represented by the 2 next to the X and Check emojis)
If the bot were to remove these duplicate emojis, it would potentially avoid any confusion that might happen.

chandler05:
## Purpose
Closes #58

### DIFF
--- a/src/events/request/RequestReactionRemovalEventHandler.ts
+++ b/src/events/request/RequestReactionRemovalEventHandler.ts
@@ -3,7 +3,7 @@ import * as log4js from 'log4js';
 import TaskScheduler from '../../tasks/TaskScheduler';
 import EventHandler from '../EventHandler';
 
-export default class RequestReactionRemovalrEventHandler implements EventHandler<'messageReactionAdd'> {
+export default class RequestReactionRemovalEventHandler implements EventHandler<'messageReactionAdd'> {
 	public readonly eventName = 'messageReactionAdd';
 
 	private logger = log4js.getLogger( 'RequestResolutionRemovalEventHandler' );
--- a/src/events/request/RequestReactionRemovalEventHandler.ts
+++ b/src/events/request/RequestReactionRemovalEventHandler.ts
@@ -12,7 +12,7 @@ export default class RequestReactionRemovalEventHandler implements EventHandler<
 		this.logger.info( `User ${ user.tag } added '${ reaction.emoji.name }' reaction to request message '${ reaction.message.id }'` );
                 const guildMember = reaction.members.fetch(user);
 
-                if ( !guildMember.roles.cache.has( #mod role id ) && !guildMember.roles.cache.has( #helper role id ) ) {
+                if ( guildMember.permissionsIn(reaction.message.channel) !== 'ADD_REACTIONS' ) {
 		     reaction.users.remove(user);
                 }
 	};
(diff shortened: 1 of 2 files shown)


src/events/request/RequestReactionRemovalEventHandler.ts
```diff
@@ -0,0 +1,19 @@
+import { MessageReaction, User } from 'discord.js';
+import * as log4js from 'log4js';
+import TaskScheduler from '../../tasks/TaskScheduler';
+import EventHandler from '../EventHandler';
+
+export default class RequestReactionRemovalrEventHandler implements EventHandler<'messageReactionAdd'> {
```
SPGoding, line 6:
```suggestion
export default class RequestReactionRemovalEventHandler implements EventHandler<'messageReactionAdd'> {
```

src/events/request/RequestReactionRemovalEventHandler.ts
```diff
@@ -0,0 +1,19 @@
...
+	public readonly eventName = 'messageReactionAdd';
+
+	private logger = log4js.getLogger( 'RequestResolutionRemovalEventHandler' );
```
SPGoding, line 9:
```suggestion
	private logger = log4js.getLogger( 'RequestReactionRemovalEventHandler' );
```

src/events/request/RequestReactionRemovalEventHandler.ts
```diff
@@ -0,0 +1,19 @@
...
+
+	// This syntax is used to ensure that `this` refers to the `RequestResolveEventHandler` object
+	public onEvent = async ( reaction: MessageReaction, user: User ): Promise<void> => {
+		this.logger.info( `User ${ user.tag } added '${ reaction.emoji.name }' reaction to request message '${ reaction.message.id }'` );
+
+		TaskScheduler.clearMessageTasks( reaction.message );
```
SPGoding, line 15: What's this line for?

SPGoding (changes requested):

src/events/request/RequestReactionRemovalEventHandler.ts
chandler05: Oops

SPGoding (approved):

SPGoding (changes requested):
Actually this is removing helpers and moderators' ability to add reactions by hand.

chandler05:
Can a mod add in the IDs for the Moderator and Helper roles?

src/events/request/RequestReactionRemovalEventHandler.ts
```diff
@@ -0,0 +1,19 @@
...
+	public readonly eventName = 'messageReactionAdd';
+
+	private logger = log4js.getLogger( 'RequestReactionRemovalEventHandler' );
+
+	// This syntax is used to ensure that `this` refers to the `RequestResolveEventHandler` object
+	public onEvent = async ( reaction: MessageReaction, user: User ): Promise<void> => {
+		this.logger.info( `User ${ user.tag } added '${ reaction.emoji.name }' reaction to request message '${ reaction.message.id }'` );
+                const guildMember = reaction.members.fetch(user);
+
+                if ( !guildMember.roles.cache.has( #mod role id ) && !guildMember.roles.cache.has( #helper role id ) ) {
```
SPGoding, line 15: It would be great to move this part to config in form of a list of strings (probably called `privilegedUsers`?). You can get the role ID with Discord developer mode enabled, but anyways, `647812320629882910` is moderator, `647812604949037056` is helper, `647812889133973514` is mojang, and `647814138818723850` is bot.

SPGoding: Another way would be to check if the user has the reaction permission, which is imo easier and better to implement.

SPGoding (approved):

PR #136, opened 2020-12-01
 Rename pending request channels with the amount of pending requests in that channel

### LINKED ISSUE #68: Rename pending request channels with amount of pending requests
It would be great if the pending request channels would be renamed so that the amount of unresolved requests is immediately visible without the need to visit the channel. (e.g. "#20-java-pending-requests")

This may be a bit difficult to realize, not only because the bot needs to know how many unresolved requests there are left (which means it also needs to track message deletions etc.) but also because there's a rate limit on how often discord bots can edit channel titles. I'm not entirely sure as for what the exact rate limits are, but they exist (the Minecraft commands discord has had trouble with the channel renaming rate limit, but they also use it very extensively).

chandler05:
## Purpose
Closes #68
## Approach
New config for channel names

### DIFF
--- a/src/BotConfig.ts
+++ b/src/BotConfig.ts
@@ -17,6 +17,7 @@ export enum PrependResponseMessageType {
 export class RequestConfig {
 	public channels: string[];
 	public internalChannels: string[];
+	public internalChannelNames: string[];
 	public logChannel: string;
 
 	public noLinkEmoji: string;
@@ -33,9 +34,13 @@ export class RequestConfig {
 	constructor() {
 		this.channels = getOrDefault( 'request.channels', [] );
 		this.internalChannels = this.channels.length ? config.get( 'request.internalChannels' ) : getOrDefault( 'request.internalChannels', [] );
+		this.internalChannelNames = this.channels.length ? config.get( 'request.internalChannelNames' ) : getOrDefault( 'request.internalChannelNames', [] );
 		this.logChannel = config.get( 'request.logChannel' );
 
 		if ( this.channels.length !== this.internalChannels.length ) {
+			throw new Error( 'There are not exactly as many Request channels and ' );
+		}
+				if ( this.channels.length !== this.internalChannelNames.length ) {
 			throw new Error( 'There are not exactly as many Request channels and ' );
 		}
 
--- a/src/BotConfig.ts
+++ b/src/BotConfig.ts
@@ -40,7 +40,7 @@ export class RequestConfig {
 		if ( this.channels.length !== this.internalChannels.length ) {
 			throw new Error( 'There are not exactly as many Request channels and ' );
 		}
-				if ( this.channels.length !== this.internalChannelNames.length ) {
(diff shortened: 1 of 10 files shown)


chandler05:
This gets the bot Rate Limited very quickly, I don't know if there is any way to do this while getting around rate limiting

violine1101 (changes requested):
A way in which this could work without catching the rate limit is by not updating the name of the channel every time a new request is posted, but rather have a task updating the channel name regularily (e.g. every 10 minutes, but that should be configurable). It might be difficult to keep track of the amount of requests in a channel though. However, the way it's implemented now, we'd probably run into rate limits constantly.
For reference, the rate limit for channel name and topic updates is 2 updates per 10 minutes per channel.

src/BotConfig.ts
```diff
@@ -35,11 +36,15 @@ export class RequestConfig {
...
 		this.channels = getOrDefault( 'request.channels', [] );
 		this.internalChannels = this.channels.length ? config.get( 'request.internalChannels' ) : getOrDefault( 'request.internalChannels', [] );
+		this.internalChannelNames = this.channels.length ? config.get( 'request.internalChannelNames' ) : getOrDefault( 'request.internalChannelNames', [] );
 		this.logChannel = config.get( 'request.logChannel' );
 
 		if ( this.channels.length !== this.internalChannels.length ) {
 			throw new Error( 'There are not exactly as many Request channels and ' );
 		}
+		if ( this.internalChannels.length !== this.internalChannelNames.length ) {
+			throw new Error( 'There are not exactly as many Internal channels and ' );
```
Jack-McKalling, line 46: ... and what?
```suggestion
			throw new Error( 'There are not exactly as many Internal channels as associated names' );
```

urielsalis:
I would keep in memory the number of requests open, and update both the pending requests and the normal requests channel(for transparency)

As for when to update, I think we should keep a timer of 1 min per channel. Every time that there is a change(ticket resolved), start the timer. Reset it back to 1 minute if there is another change. After the 1 min update both channel names

violine1101:
> I would keep in memory the number of requests open, and update both the pending requests and the normal requests channel(for transparency)

For the normal channels, I think it makes more sense to update the channel _topic_ for the public channels. But yeah, it would be nice to have that count public too.

> As for when to update, I think we should keep a timer of 1 min per channel. Every time that there is a change(ticket resolved), start the timer. Reset it back to 1 minute if there is another change. After the 1 min update both channel names

This will just cause the bot to get rate limited by Discord.

I'm probably going to open a completely new PR for this, so I'm going to copy these considerations to the issue.

urielsalis:
> This will just cause the bot to get rate limited by Discord, or alternatively, the channel name to never be updated.

If its only updated 1 minute after the last change, I dont think it will get rate limited. Way more chances that every 10 mins too

violine1101:
The rate limit is 2 channel renames per 10 minutes, as I said. In the worst case with your method the channel can be renamed up to 9 times in that time frame.

PR #139, opened 2020-12-04
Don't allow requests for tickets younger than 24 hours

chandler05:
## Purpose
Stop allowing requests for brand new tickets.
I believe this is still pending internal discussion/resolution, but the code is ready whenever it needs to be merged

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -23,8 +23,8 @@ projects:
   - WEB
 
 request:
-  noLinkEmoji: ⛔
-  noLinkWarningLifetime: 5000
+  warningEmoji: ⛔
+  warningLifetime: 5000
   waitingEmoji: ⌛
 
   invalidRequestJql: created < 24h
--- a/config/default.yml
+++ b/config/default.yml
@@ -28,7 +28,7 @@ request:
   warningLifetime: 5000
   waitingEmoji: ⌛
 
-  invalidRequestJql: created < 24h
+  invalidRequestJql: created > -24h
 
   suggestedEmoji:
     - ✅
--- a/config/default.yml
+++ b/config/default.yml
@@ -23,7 +23,8 @@ projects:
   - WEB
 
 request:
-  warningEmoji: ⛔
+  invalidTicketEmoji: ⏸
+  noLinkEmoji: ⛔
   warningLifetime: 5000
   waitingEmoji: ⌛
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -27,6 +27,8 @@ request:
   noLinkWarningLifetime: 5000
   waitingEmoji: ⌛
 
+  invalidRequestJql: created < 24h
+
   suggestedEmoji:
     - ✅
     - ☑️
--- a/src/commands/MentionCommand.ts
+++ b/src/commands/MentionCommand.ts
@@ -8,11 +8,17 @@ export default class MentionCommand extends Command {
 		return `(?:${ BotConfig.projects.join( '|' ) })-\\d+`;
 	}
 
-	public static get ticketIdRegex(): RegExp {
+	/**
+	 * @returns A NEW regex object every time. You have to store it as a variable if you use `exec` on it, otherwise you will encounter infinite loops.
+	 */
+	public static getTicketIdRegex(): RegExp {
 		return new RegExp( `(?<=^|[^${ BotConfig.forbiddenTicketPrefix }])(?<=${ BotConfig.requiredTicketPrefix })(${ MentionCommand.ticketPattern })`, 'g' );
 	}
 
-	public static get ticketLinkRegex(): RegExp {
+	/**
+	 * @returns A NEW regex object every time. You have to store it as a variable if you use `exec` on it, otherwise you will encounter infinite loops.
+	 */
+	public static getTicketLinkRegex(): RegExp {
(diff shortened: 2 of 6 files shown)


src/events/request/RequestEventHandler.ts
```diff
@@ -58,6 +60,33 @@ export default class RequestEventHandler implements EventHandler<'message'> {
 			return;
 		}
 
+		if ( BotConfig.request.invalidRequestJql ) {
+			const searchResults = await this.jira.search.search ( {
+				jql: BotConfig.request.invalidRequestJql,
+				fields: ['key'],
+			} );
```
SPGoding, line 67: How about changing the JQL to `` `(${ BotConfig.request.invalidRequestJql }) AND key in (${ tickets.join( ',' ) })``` and check if the result is empty so that we are not querying thousands of tickets everytime?

chandler05: Hehe good point

src/events/request/RequestEventHandler.ts
```diff
@@ -58,6 +60,32 @@ export default class RequestEventHandler implements EventHandler<'message'> {
...
+				try {
+					await origin.react( BotConfig.request.noLinkEmoji );
+				} catch ( error ) {
+					this.logger.error( error );
+				}
+
+				try {
+					const warning = await origin.channel.send( `${ origin.author }, your request (<${ origin.url }>) contains a ticket that is less than 24 hours old. Please wait until it is at least one day old before making a request.` );
+
+					const timeout = BotConfig.request.noLinkWarningLifetime;
```
SPGoding, line 80: As we're using `noLinkWarningLifetime` here too, let's rename it to something like `requestWarningLifetime` :eyes:

chandler05: K, I'll change the emoji's name as well

SPGoding (commented):
Actually as suggested somewhere in Discord, using the pause emoji (⏸) might be better for this case. So we might want to split `warningEmoji`

config/default.yml
```diff
@@ -23,10 +23,12 @@ projects:
   - WEB
 
 request:
-  noLinkEmoji: ⛔
-  noLinkWarningLifetime: 5000
+  warningEmoji: ⛔
+  warningLifetime: 5000
   waitingEmoji: ⌛
 
+  invalidRequestJql: created < 24h
```
SPGoding, line 30:
```suggestion
  invalidRequestJql: created > -24h
```

SPGoding (approved):
image

src/commands/MentionCommand.ts
```diff
@@ -8,11 +8,17 @@ export default class MentionCommand extends Command {
 		return `(?:${ BotConfig.projects.join( '|' ) })-\\d+`;
 	}
 
-	public static get ticketIdRegex(): RegExp {
+	/**
+	 * @returns A NEW regex object every time. You have to store it as a variable if you use `exec` on it, otherwise you will encounter infinite loops.
+	 */
+	public static getTicketIdRegex(): RegExp {
```
violine1101, line 14: we should probably extract these functions into a util class at some point

src/events/request/RequestEventHandler.ts
```diff
@@ -58,6 +65,32 @@ export default class RequestEventHandler implements EventHandler<'message'> {
 			return;
 		}
 
+		if ( BotConfig.request.invalidRequestJql ) {
+			const tickets = Array.from( this.getTickets( this.replaceTicketReferencesWithRichLinks( origin.content, regex ) ) );
```
violine1101, line 69: What's the reason for the `replaceTicketReferencesWithRichLinks` here? Wouldn't it work without it too?

src/events/request/RequestEventHandler.ts
```diff
@@ -91,6 +124,16 @@ export default class RequestEventHandler implements EventHandler<'message'> {
 		}
 	};
 
+	private getTickets( content: string ): Set<string> {
```
violine1101, line 127: Why does this return a set, not an array? It will be converted into an array anyway.

violine1101 (commented):
LGTM, only the `getTickets` call looks a bit iffy to me.

src/events/request/RequestEventHandler.ts
chandler05: Idk why I used that, it should work just fine without it

src/events/request/RequestEventHandler.ts
```diff
@@ -91,6 +124,16 @@ export default class RequestEventHandler implements EventHandler<'message'> {
...
 	};
 
+	private getTickets( content: string ): string[] {
+		let ticketMatch: RegExpExecArray;
+		const regex = MentionCommand.getTicketIdRegex();
+		const ticketMatches: Set<string> = new Set();
+		while ( ( ticketMatch = regex.exec( content ) ) !== null ) {
+			ticketMatches.add( ticketMatch[1] );
+		}
+		return Array.from( ticketMatches );
```
violine1101, line 134: There's no reason to use a set here.
```suggestion
		const ticketMatches: string[] = [];
		while ( ( ticketMatch = regex.exec( content ) ) !== null ) {
			ticketMatches.push( ticketMatch[1] );
		}
		return ticketMatches;
```

violine1101 (approved):

PR #140, opened 2020-12-05
Enable bulk action by Jira filter

chandler05:
## Purpose
Large amounts of requests like adding versions can end up being very tedious to do Individually. With a new reaction and command, the bot will provide a Jira filter will all of the tickets from specific requests in order to use a Bulk Action.
## Approach
Workflow:
1. Add the bulk reaction to any requests
2. Type '!jira bulk'
3. The bot will send a Jira link with a filter containing all tickets in the reacted to requests
4. The bot will remove the reactions

### DIFF
--- a/src/commands/BulkCommand.ts
+++ b/src/commands/BulkCommand.ts
@@ -54,7 +54,7 @@ export default class BulkCommand extends PrefixCommand {
 			Command.logger.error( err );
 			return false;
 		}
-		const filter = `https://bugs.mojang.com/browse/${ firstMentioned }?jql=key%20in(${ ticketKeys.join( '%2C' ) })`
+		const filter = `https://bugs.mojang.com/browse/${ firstMentioned }?jql=key%20in(${ ticketKeys.join( '%2C' ) })`;
 		try {
 			await message.channel.send( `${ message.author.toString() } ${ filter }` );
 		} catch {
--- a/src/commands/BulkCommand.ts
+++ b/src/commands/BulkCommand.ts
@@ -20,7 +20,7 @@ export default class BulkCommand extends PrefixCommand {
 			try {
 				const internalChannel = await DiscordUtil.getChannel( internalChannelId );
 				if ( internalChannel instanceof TextChannel ) {
-					const channelMessages = internalChannel.messages;
+					const channelMessages = internalChannel.messages.cache;
 					for await ( const channelMessage of channelMessages ) {
 						const reaction = channelMessage.reactions.cache.get( BotConfig.request.bulkEmoji );
 						if ( reaction.users.cache.get( message.author ) ) {
--- a/src/commands/BulkCommand.ts
+++ b/src/commands/BulkCommand.ts
@@ -13,7 +13,7 @@ export default class BulkCommand extends PrefixCommand {
 			return false;
 		}
 
-		let bulkMessages: Messages[];
+		let bulkMessages: Message[];
 
 		for ( let i = 0; i < BotConfig.request.internalChannels.length; i++ ) {
 			const internalChannelId = BotConfig.request.internalChannels[i];
--- a/src/commands/BulkCommand.ts
+++ b/src/commands/BulkCommand.ts
@@ -32,7 +32,7 @@ export default class BulkCommand extends PrefixCommand {
 				Command.logger.error( err );
 				return false;
 			}
-	        }
+		}
 
 		let firstMentioned: string;
 		let ticketKeys: string[];
--- a/src/commands/BulkCommand.ts
+++ b/src/commands/BulkCommand.ts
@@ -21,7 +21,7 @@ export default class BulkCommand extends PrefixCommand {
 				const internalChannel = await DiscordUtil.getChannel( internalChannelId );
 				if ( internalChannel instanceof TextChannel ) {
 					const channelMessages = internalChannel.messages;
-					for ( const channelMessage of channelMessages ) {
+					for await ( const channelMessage of channelMessages ) {
 						const reaction = channelMessage.reactions.cache.get( BotConfig.request.bulkEmoji );
 						if ( reaction.users.cache.get( message.author ) ) {
 							bulkMessages.add( channelMessage );
--- a/src/commands/BulkCommand.ts
+++ b/src/commands/BulkCommand.ts
@@ -13,7 +13,7 @@ export default class BulkCommand extends PrefixCommand {
 			return false;
 		}
 
-		let bulkMessages: Set<Message>;
+		let bulkMessages: Messages[];
 
 		for ( let i = 0; i < BotConfig.request.internalChannels.length; i++ ) {
 			const internalChannelId = BotConfig.request.internalChannels[i];
@@ -24,7 +24,7 @@ export default class BulkCommand extends PrefixCommand {
 					for await ( const channelMessage of channelMessages ) {
 						const reaction = channelMessage.reactions.cache.get( BotConfig.request.bulkEmoji );
 						if ( reaction.users.cache.get( message.author.id ) ) {
-							bulkMessages.add( channelMessage );
+							bulkMessages.push( channelMessage );
 						}
 					}
 				}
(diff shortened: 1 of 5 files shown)


SPGoding (commented):
From what I understand, the bulk emoji will result in showing the prompt and solving it after 10 seconds as there's no special check in `RequestEventHandler`.

Also, how about adding tickets IDs to an existing `Set` as soon as someone reacted an request with the `bulkEmoji`, so we're not iterating the whole internal channel when executing `!jira bulk` command?

chandler05:
Wouldn't that stop the filters from being individual for each person?

SPGoding:
It's still possible to make a `Map` from user tags to `Set`s :eyes:


Awwwso, instead of asking the volunteer to execute `!jira bulk` manually to see the filter, how about adding it to a new field of internal embed in some format like `[bulk filter](https://bugs.mojang.com/jql@#%$^^%)` as soon as the volunteer reacted it with the `bulkEmoji`? This way, the `Set` (or `Map`) maintained in `RequestEventHandler` doesn't have to be exposed to another handler.

chandler05:
Where would that be posted?

SPGoding:
image

chandler05:
The filter is supposed to extend over any request with the reaction. That confuses me as to how that would work. Plus, wouldn't that only work for one user?

SPGoding:
Yeah I just gave it another thought and it's quite difficult to maintain if we put the JQL link directly in the embed. So I guess we still have to go with the `!jira bulk` command 🤔

chandler05:
I changed the code to make it so a map of users to bulk messages is stored instead of iterating through all of the messages in the channel each time. I also fixed the bulk emoji from triggering the prompt/resolution.

chandler05:
Adding an emoji after '!jira bulk' should now resolve all of the requests in the bulk filter with that emoji

src/events/request/RequestBulkRemoveEventHandler.ts
```diff
@@ -0,0 +1,27 @@
...
+	public onEvent = async ( reaction: MessageReaction, user: User ): Promise<void> => {
+		this.logger.info( `User ${ user.tag } removed '${ reaction.emoji.name }' reaction from request message '${ reaction.message.id }'` );
+
+		if ( BulkCommand.currentBulkReactions.has( user ) ) {
+			const currentMessages = BulkCommand.currentBulkReactions.get( user );
+			for ( let i = 0; i <= currentMessages.length; i++ ) {
+				const currentMessage = currentMessages[i];
+				if ( currentMessage === reaction.message ) {
+					BulkCommand.currentBulkReactions.delete( user );
+					BulkCommand.currentBulkReactions.set( user, currentMessages.splice( i, 1 ) );
```
SPGoding, line 21: `splice` returns the deleted elements, not the array itself. Also, `currentMessages` is a reference to the array which isn't reassigned anywhere, so all we need to do is a simple
```suggestion
					currentMessages.splice( i, 1 );
```

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+				return false;
+			}
+		}
+
+		let ticketKeys: string[];
+		let firstMentioned: string;
+
+		try {
+			const bulkMessages = BulkCommand.currentBulkReactions.get( message.author );
+			let originMessages: Message[];
```
SPGoding, line 39: This is never initialized.

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+		if ( args.length ) {
+			const customEmoji = /^<a?:(\w+):(\d+)>/;
+			const unicodeEmoji = emojiRegex();
+
+			if ( customEmoji.test( args ) || unicodeEmoji.test( args ) ) {
+				rawEmoji = args;
+				const emojiMatch = customEmoji.exec( args );
+				if ( emojiMatch ) {
+					rawEmoji = emojiMatch[2];
+				}
```
SPGoding, line 27: The logic here looks tricky. I'd suggest using

```suggestion
			if ( unicodeEmoji.test( args ) ) {
				rawEmoji = args;
			} else if ( customEmoji.test( args ) ) {
				rawEmoji = customEmoji.exec( args )[2];
```

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+			}
+		}
+
+		let ticketKeys: string[];
+		let firstMentioned: string;
+
+		try {
+			const bulkMessages = BulkCommand.currentBulkReactions.get( message.author );
+			let originMessages: Message[];
+			for ( const bulk of bulkMessages ) {
```
SPGoding, line 40: `bulkMessages` could be `undefined`.

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+
+	private getTickets( content: string ): string[] {
+		let ticketMatch: RegExpExecArray;
+		const regex = MentionCommand.getTicketIdRegex();
+		const ticketMatches: string[] = [];
+		while ( ( ticketMatch = regex.exec( content ) ) !== null ) {
+			ticketMatches.push( ticketMatch[1] );
+		}
+		return ticketMatches;
+	}
```
SPGoding, line 77: I'd suggest moving both this and the one in `RequestEventHandler` to `RequestsUtil`.

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+				if ( emojiMatch ) {
+					rawEmoji = emojiMatch[2];
+				}
+			} else {
+				await message.channel.send( `**Error:** ${ args } is not a valid emoji.` );
+				return false;
+			}
+		}
+
+		let ticketKeys: string[];
```
SPGoding, line 34: This is never initialized.

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+		let firstMentioned: string;
+
+		try {
+			const bulkMessages = BulkCommand.currentBulkReactions.get( message.author );
+			let originMessages: Message[];
+			for ( const bulk of bulkMessages ) {
+				originMessages.push( await RequestsUtil.getOriginMessage( bulk ) );
+				await bulk.reactions.cache.get( BotConfig.request.bulkEmoji ).users.remove( message.author );
+			}
+			originMessages.forEach( origin => this.getTickets( origin.content ).forEach( ticket => ticketKeys.push( ticket ) ) );
```
SPGoding, line 44: The nested `forEach` here is confusing.
```suggestion
			originMessages.forEach( origin => ticketKeys.push( ...this.getTickets( origin.content ) ) );
```

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+			firstMentioned = ticketKeys[0];
+			if ( rawEmoji ) {
+				bulkMessages.forEach( resolvable => TaskScheduler.addOneTimeMessageTask(
+					resolvable,
+					new ResolveRequestMessageTask( rawEmoji, message.author ),
+					BotConfig.request.resolveDelay || 0
+				) );
```
SPGoding, line 51: I'd suggest delegating this operation to `RequestResolveEventHandler` instead of copying this piece of code here.

src/commands/BulkCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+
+		const filter = `https://bugs.mojang.com/browse/${ firstMentioned }?jql=key%20in(${ ticketKeys.join( '%2C' ) })`;
+
+		try {
+			await message.channel.send( `${ message.author.toString() } ${ filter }` );
+		} catch {
+			return false;
+		}
+
+		return true;
```
SPGoding, line 66: I think it would be fine to delete the `!jira bulk` command somewhere. And also delete the bot's own filter message after some time.

src/commands/BulkCommand.ts
chandler05: The only problem is there is no actual message reaction to handle at this point, just the emoji

chandler05:
The feature is working very well, but for some reason it seems to ocassionally miss some of the bulk emoji reactions in the internal channels

PR #144, opened 2020-12-05
Fix formatting errors

chandler05:


SPGoding (approved):
Btw I recommend install the `dbaeumer.vscode-eslint` extension :eyes:

PR #145, opened 2020-12-07
Enforce testing request channel rules

chandler05:
## Purpose
Enforce the following rules in the testing request channels:
1. No newly created tickets
2. Request must have exactly one link


SPGoding:
> Request must have exactly one link

Why?

src/util/RequestsUtil.ts
```diff
@@ -93,4 +95,32 @@ export class RequestsUtil {
...
+				return true;
+			}
+		} catch ( err ) {
+			this.logger.error( err );
+			return false;
+		}
+	}
+
+	public static checkTicketLinks( content: string ): number {
+		const regex = new RegExp( `(?:${ MentionCommand.getTicketLinkRegex().source }|(${ MentionCommand.ticketPattern }))(\\?\\S+)?`, 'g' );
```
violine1101, line 118: The `MentionCommand` regexes should probably be extracted into their own util class. But out of scope for this PR.

src/events/request/RequestEventHandler.ts
```diff
@@ -24,94 +23,135 @@ export default class RequestEventHandler implements EventHandler<'message'> {
...
-		}
-
-		try {
-			await origin.reactions.removeAll();
-		} catch ( error ) {
-			this.logger.error( error );
-		}
-
-		const regex = new RegExp( `(?:${ MentionCommand.getTicketLinkRegex().source }|(${ MentionCommand.ticketPattern }))(\\?\\S+)?`, 'g' );
+		if ( !BotConfig.request.testingRequestChannels || !BotConfig.request.testingRequestChannels.includes( origin.channel.id ) ) {
```
violine1101, line 26: I don't like this large if statement that wraps around essentially the whole function.

src/events/request/RequestEventHandler.ts
```diff
@@ -24,94 +23,135 @@ export default class RequestEventHandler implements EventHandler<'message'> {
...
-				const timeout = BotConfig.request.warningLifetime;
-				await warning.delete( { timeout } );
+				await origin.reactions.removeAll();
 			} catch ( error ) {
 				this.logger.error( error );
 			}
 
-			return;
-		}
+			const regex = new RegExp( `(?:${ MentionCommand.getTicketLinkRegex().source }|(${ MentionCommand.ticketPattern }))(\\?\\S+)?`, 'g' );
```
violine1101, line 41: This regex is now far away from where it's actually used; and it probably can get removed altogether (as it was replaced with the util function).

src/util/RequestsUtil.ts
```diff
@@ -93,4 +95,32 @@ export class RequestsUtil {
...
+			} else {
+				return true;
+			}
+		} catch ( err ) {
+			this.logger.error( err );
+			return false;
+		}
+	}
+
+	public static checkTicketLinks( content: string ): number {
```
violine1101, line 117: I wonder if we actually need this function at all, we already have `RequestEventHandler#getTickets` which does essentially the same thing already but better (though arguably more confusing).

src/util/RequestsUtil.ts
```diff
@@ -93,4 +95,32 @@ export class RequestsUtil {
 		}
 		return this.hashCode( resolver.tag ) & 0x00FFFFFF;
 	}
+
+	public static async checkTicketValidity( ticketKeyString: string ): Promise<boolean> {
+		try {
+			const searchResults = await MojiraBot.jira.issueSearch.searchForIssuesUsingJqlGet( {
+				jql: `(${ BotConfig.request.invalidRequestJql }) AND key in (${ ticketKeyString })`,
```
violine1101, line 102: Would be nice to have different invalid JQLs for different channels; I might come back to that later. Out of scope for this PR though.

violine1101 (commented):
Looks good; however there are a few minor things that I don't like. Not a big thing though and I'm probably going to fix them myself to finally get this PR merged.

src/events/request/RequestEventHandler.ts
```diff
@@ -24,94 +23,135 @@ export default class RequestEventHandler implements EventHandler<'message'> {
...
-				? RequestsUtil.getResponseMessage( origin )
-				: '';
+				const copy = await internalChannel.send( response, embed ) as Message;
 
-			const copy = await internalChannel.send( response, embed ) as Message;
+				if ( BotConfig.request.suggestedEmoji ) {
+					await ReactionsUtil.reactToMessage( copy, [...BotConfig.request.suggestedEmoji] );
+				}
+			}
+		} else if ( !origin.guild.member( origin.author ).permissionsIn( await DiscordUtil.getChannel( BotConfig.request.logChannel ) ).has( 'VIEW_CHANNEL' ) ) {
```
violine1101, line 112: Oh, forgot to mention. This is a really smart way to query who's exempt from this rule. Might extract that one into an utiil function and use elsewhere too in the future. (Although it's probably more clean to have a list of roles...)

violine1101 (approved):

PR #147, opened 2020-12-09
Add command for sending messages as MojiraBot

### LINKED ISSUE #112: Add a new '!jira send' command
## The Problem
Some people accuse the bot of being dumb. The bot needs to reply to them to prove that they are right.

And we might want to use embed sometimes (e.g. sending news with rich links).

## Possible Solution
Adding a new `!jira send <channel> (plain <message>|embed <embed description>)` command (or with some other syntax). By executing this command we could send rich messages in other channels (e.g. `#news`) ~~and fight back when someone curses the bot~~.

chandler05:
## Purpose
Closes #112


violine1101 (approved):
Looks good! :+1:

I'll add one minor change that I'll commit myself in a moment: It should be transparent who sent the message; whoever sent the command should be able to be held accountable for messages that they send via the bot. Thus I removed the auto delete thing and just added a reaction. Since it's possible to specify channel that's not a big deal.

src/commands/SendCommand.ts
```diff
@@ -0,0 +1,91 @@
+import { Message, MessageEmbed, TextChannel, DMChannel, NewsChannel } from 'discord.js';
+import PrefixCommand from './PrefixCommand';
+import PermissionRegistry from '../permissions/PermissionRegistry';
+import Command from './Command';
+
+export default class ShutdownCommand extends PrefixCommand {
```
UserTeemu, line 6: I'm not very experienced in TypeScript, but this class being called `ShutdownCommand` seems a bit odd, when the file is `SendCommand` and `ShutdownCommand` is a completely separate command.

chandler05: You are correct, this is not the first time I've done this 😆

PR #149, opened 2020-12-26
Fix embeds not sending if title is too long

dericksonmark:
## Purpose
Jira allows summaries to be up to 255 characters long. Discord has this same limit for embed titles. However, when creating an embed, MojiraBot takes not only the summary, but the ticket ID as well, and puts them in the title. This causes all embeds of tickets with summaries longer than 246 characters (with a 6-digit ID) to fail to be sent.
## Approach
If the embed title length (including the ID and the summary) exceeds 251 characters, then it will be shortened and an ellipsis will be added to the end.

I tried my best to make working code. It does work, but it may not be ideal.
## Results
Screen Shot 2020-12-27 at 10 57 27 AM
The embed is correctly sent.

### DIFF
--- a/src/mentions/SingleMention.ts
+++ b/src/mentions/SingleMention.ts
@@ -93,7 +93,7 @@ export class SingleMention extends Mention {
 
 		const embed = new MessageEmbed();
 		embed.setAuthor( ticketResult.fields.reporter.displayName, ticketResult.fields.reporter.avatarUrls['48x48'], 'https://bugs.mojang.com/secure/ViewProfile.jspa?name=' + encodeURIComponent( ticketResult.fields.reporter.name ) )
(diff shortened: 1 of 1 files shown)


src/mentions/SingleMention.ts
```diff
@@ -151,4 +151,11 @@ export class SingleMention extends Mention {
 
 		return embed;
 	}
+	private shorten( input: string ): string {
```
SPGoding, line 154: Some name like `ensureLength` might be more descriptive.

dericksonmark: I have changed this, thanks!

chandler05 (approved):

violine1101 (approved):

PR #155, opened 2021-01-08
Update role selection messages upon restart

violine1101:
## Purpose
Automatically update all role selection messages when needed

## Approach
Edit the role selection creation routine so that it compares previous and current settings, and if they aren't equal, edits the existing message

Additionally:
* Added "color" setting for role selection messages
* Added description for both role selection messages, as well as singular roles
* Removed `!jira roles` command as it is no longer needed
* Fixed a minor bug with radio role selection mode (still doesn't work quite right though)

## Future work
N/A


SPGoding (approved):
This works really well!

PR #178, opened 2021-03-25
Fix role selection channel mention in tips command

chandler05:


SPGoding (approved):

dericksonmark (approved):

SPGoding (commented):
I was about to say that it might be better to store the channel ID in config, but we were already including the guild-specific `#role-selection` before anyways so I guess we're fine here ;D

PR #180, opened 2021-03-25
Switch help and tips commands to embeds

chandler05:
## Purpose
Currently the command text looks weird and spammy, which can be fixed by moving the command response into an embed. This also changes the formatting of the message slightly.


violine1101:
While we're at it, might make sense to add the person who triggered the command to the embed too, similar to bug report mentions. Currently it's not possible to tell who did that, since the invocation command is removed immediately.

chandler05:
I was originally planning on doing that, but I guess I forgot :). Done

SPGoding (approved):

PR #183, opened 2021-04-08
Add default value for deployment in start.sh

SunCatMC:
## Purpose
To be able to do restart the bot via a cron job

## Approach
Add variable $DEPLOYMENT that is set to $1 when it exists, and to "main" if not

## Future work
Test this on an actual bot/server, but bash was tested with echo

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,2 +1,4 @@
+# Make it so the default deployment is "main"
+DEPLOYMENT=${1:-main}
 # Start a screen called "mojiradiscordbot-<deployment>" with the node process
-/usr/bin/screen -S mojiradiscordbot-$1 -d -m bash -c "NODE_ENV=$1 node bin" ; exit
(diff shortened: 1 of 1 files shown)


urielsalis (approved):

violine1101:
Uhm. So I'm not quite happy about the default param being `main`. If I have the bot running locally, I want it to be `beta`. Which is why no default was specified before. I don't want to accidentally launch a second instance of the bot that might interfere with the currently running one.

I'm not quite sure what the best way to fix this would have been. It would have been better if we'd first have had an issue where we could have discussed this.

violine1101:
For example, a potential solution would have been to throw an error if `$1` is not specified, and have a separate `start.sh` file in the super folder that's not in VCS for the cronjob to run, which would just contain `./mojira-discord-bot/start.sh main`.

PR #190, opened 2021-05-05
Ability to add progress/comments to internal requests

### LINKED ISSUE #29: Allow to add links / progress indicators to internal requests
I don't know how doable this one is, but it would be cool to have some way of allowing to add messages to internal requests, for example using a command.
This would be useful if a request needs an internal discussion to show other volunteers that the request is already in progress (since reacting with an hourglass isn't an option anymore).

Example of a usecase:
image

chandler05:
## Purpose
Close #29 
## Approach
Sending a message in a pending request channel with the following syntax will add it to the referenced message:
`<message id (not message link)> <Message>`

It will be attached to the request in the following syntax:
internal-request - Discord 5_5_2021 3_12_02 PM (2)

This also changes how request updates are handled. Instead of deleting the internal message, it will edit it in order to retain any progress or comments.
## Future work
Currently does not allow deletion or editing the comments, but I don't think we need this right now.

### DIFF
--- a/src/events/internal/InternalProgressEventHandler.ts
+++ b/src/events/internal/InternalProgressEventHandler.ts
@@ -1,0 +1,28 @@
+import { Message, MessageReaction, User } from 'discord.js';
+import * as log4js from 'log4js';
+import BotConfig, { PrependResponseMessageType } from '../../BotConfig';
+import MojiraBot from '../../MojiraBot';
+import AddProgressMessageTask from '../../tasks/AddProgressMessageTask';
+import TaskScheduler from '../../tasks/TaskScheduler';
+import EventHandler from '../EventHandler';
+
+export default class InternalProgressEventHandler implements EventHandler<'message'> {
+	public readonly eventName = 'message';
+
+	private logger = log4js.getLogger( 'InternalProgressEventHandler' );
+
+	// This syntax is used to ensure that `this` refers to the `RequestResolveEventHandler` object
+	public onEvent = async ( origin: Message ): Promise<void> => {
+		const messageId = origin.content.split( /\s/ )[0];
+		try {
+			const progressedRequest = (await origin.channel.messages.fetch( messageId ));
+			TaskScheduler.addOneTimeMessageTask(
+				origin,
+				new AddProgressMessageTask( progressedRequest ),
+				BotConfig.request.progressMessageAddDelay || 0
+			);
+		} catch ( err ) {
+			this.logger.error( err );
+		}
+	};
+}
--- a/src/events/internal/InternalProgressEventHandler.ts
+++ b/src/events/internal/InternalProgressEventHandler.ts
@@ -16,9 +16,9 @@ export default class InternalProgressEventHandler implements EventHandler<'messa
 		if ( !messageId.match( /[0-9]{18}/ ) ) {
 			try {
 				const error = await origin.channel.send( `${ origin.author.toString() } ${ messageId } is not a valid message ID!` );
-				
+
 				const timeout = BotConfig.request.warningLifetime;
-				
+
 				await error.delete( { timeout } );
 			} catch ( err ) {
 				this.logger.error( err );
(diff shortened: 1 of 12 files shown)


Jack-McKalling:
Regarding the deletion of comments, what would happen if the request was closed and immediately reopened? Would that "unlink" the comments/progress? Although if it would, it would only allow clearing all at once.

chandler05:
Yes that would remove all comments.

src/events/internal/InternalProgressEventHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+				const timeout = BotConfig.request.warningLifetime;
+				
+				await error.delete( { timeout } );
+			} catch ( err ) {
+				this.logger.error( err );
+			}
+			return;
+		}
+		try {
+			const progressedRequest = (await origin.channel.messages.fetch( messageId ));
```
violine1101, line 29: If the message ID doesn't exist, an error is thrown and nothing happens. Would be good to signal this error to the user too.

src/tasks/AddProgressMessageTask.ts
```diff
@@ -0,0 +1,41 @@
...
+				await origin.delete();
+			} catch ( error ) {
+				AddProgressMessageTask.logger.error( error );
+			}
+		}
+
+		if ( comment ) {
+			try {
+				const embed = this.request.embeds[0];
+				embed.addField( date.toDateString(), `${ user.tag } - ${ comment.replace( `${ this.request.id } `, '' ).replace( `${ this.request.id }\n`, '' ) }` );
```
violine1101, line 34: I'd imagine that instead of `user.tag` we could also directly use `user` here for a nice link thingy. But not too important.

src/events/request/RequestUpdateEventHandler.ts
```diff
@@ -0,0 +1,46 @@
+import { Message, TextChannel } from 'discord.js';
+import * as log4js from 'log4js';
+import EventHandler from '../EventHandler';
+import { RequestsUtil } from '../../util/RequestsUtil';
+import DiscordUtil from '../../util/DiscordUtil';
+
+export default class RequestUpdateEventHandler implements EventHandler<'messageUpdate'> {
```
violine1101, line 7: Yay! It's always somewhat bothered me that "update" was just "delete and post anew". This is way better :+1:

chandler05: Thanks! :D

src/tasks/AddProgressMessageTask.ts
chandler05: Done

src/events/internal/InternalProgressEventHandler.ts
chandler05: Done

src/events/request/RequestUpdateEventHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+			for ( const [, internalMessage] of internalChannel.messages.cache ) {
+				const result = await RequestsUtil.getOriginIds( internalMessage );
+				if ( !result ) {
+					continue;
+				}
+				if ( result.channelId === oldMessage.channel.id && result.messageId === oldMessage.id ) {
+					try {
+						const embed = internalMessage.embeds[0];
+						embed.setDescription( RequestsUtil.getRequestDescription( newMessage ) )
+						await internalMessage.edit( embed );
```
violine1101, line 38: My IDE is screaming here because the semicolon is missing (in other places too). Running `npm run lint` doesn't find these issues though (and thus CI succeeds). Seems like eslint is still broken?! Sigh. Need to look into that.

(You don't need to do anything here, I've got a commit ready)

chandler05: Sounds like I need to change some settings on my IDE as well, I didn't even realize that the semicolon was missing and my IDE didn't care either :/

violine1101 (approved):
Looks good! Pretty barebones but that's fine cuz it's internal only.

In the future might make sense to make a command out of this, but I'd rather wait with that till slash commands are ready. It's fine as it is now.

Also :+1: for the live fixing while I'm writing the review haha

Issue #194, opened 2021-05-11
Progress comments are still added to internal request messages if the comment is deleted


dericksonmark:
## The Bug
If a helper or moderator deletes their progress comment to an internal request message, the message is still added to the request embed.
## Expected behavior
MojiraBot should not add the message to the request if the message was deleted.

PR #195, opened 2021-05-12
Fix check for deleted message

dericksonmark:
## Purpose
Fixing #194.
## Approach
The problem was that `origin === undefined` didn't always work correctly, and there needed to be another check to see if the message was deleted. The check was changed to `origin === undefined || origin.deleted`.


chandler05 (approved):

chandler05:
(Build never ran on this PR due to being from a forked repo, ran the build on a separate branch successfully)

PR #196, opened 2021-05-12
Replace internal request footer time with request's original created time

### LINKED ISSUE #185: Add time of the request itself to internal request log
## The Problem
The internal request log currently doesn't include date and time of the request itself, even though those could be useful.

## Possible Solution
Add another field for "Date/time" to the log messages.

chandler05:
## Purpose
Close #185
## Approach
I see no reason to keep the date of the message's creation in the footer, so I changed it to the request's original creation date.


violine1101 (approved):

SPGoding:
Wanted to say that after this change the time in the internal message can no longer reflect edit time, but that doesn't happen often and I don't see its use case anyways.

PR #197, opened 2021-05-12
Replace MarkdownUtil.escape with discord.js escapeMarkdown function

### LINKED ISSUE #184: Incorrect markdown escape in multiple-tickets rich embed
## The Bug

A multiple-tickets rich embed with a title that includes markdown has visible markdown escapes.

Introduced by #160

This only occurs on my phone(Android), not on my laptop(Windows). Looks like Android(my Android phone) does not support markdown in hyperlinks in a rich embed description while Windows does. This means it is not possible to completely fix this with a change to the bot. Best solution would be to only add markdown escapes to valid markdown, instead of all markdown symbols like currently.

## Example

Screenshot_20210412-084738_Discord
^ this would not be valid markdown, yet backslashes are still added. 

## Expected behavior

No visible markdown escapes or markdown escapes only added to valid markdown which will decrease the cases of this issue occuring.

chandler05:
## Purpose
Fix #184
## Approach
Use discord.js's provided function `escapeMarkdown` to replace `MarkdownUtil.escape`

### DIFF
--- a/src/commands/SearchCommand.ts
+++ b/src/commands/SearchCommand.ts
@@ -1,8 +1,7 @@
-import { Message, MessageEmbed } from 'discord.js';
+import { Message, MessageEmbed, Util } from 'discord.js';
 import PrefixCommand from './PrefixCommand';
 import BotConfig from '../BotConfig';
 import MojiraBot from '../MojiraBot';
-import { MarkdownUtil } from '../util/MarkdownUtil';
 
 export default class SearchCommand extends PrefixCommand {
 	public readonly aliases = ['search', 'find'];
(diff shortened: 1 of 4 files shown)


violine1101 (approved):
Thank you! :+1:

This doesn't actually fix #184 quite yet, though. It seems like inside of links you can't have any formatting anyway; so the escaping in there is unnecessary and causes that bug on mobile. Only MultipleMention is affected. I'll apply that change and everything is good \o/

violine1101:
Actually, seems like the new search command is affected too

violine1101:
Before: 


After:

PR #198, opened 2021-05-12
Ignore lines with quotes in ticket mentions

### LINKED ISSUE #168: Ignore ticket mentions in quotes
## The Problem
It often happens that you quote someone but don't want to create the embed.

## Possible Solution
If a line starts with `>`, ignore all mentions in it.

chandler05:
## Purpose
Fix #168

### DIFF
--- a/src/commands/MentionCommand.ts
+++ b/src/commands/MentionCommand.ts
@@ -12,7 +12,7 @@ export default class MentionCommand extends Command {
 	 * @returns A NEW regex object every time. You have to store it as a variable if you use `exec` on it, otherwise you will encounter infinite loops.
 	 */
 	public static getTicketIdRegex(): RegExp {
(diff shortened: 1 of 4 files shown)


src/commands/MentionCommand.ts
```diff
@@ -32,6 +32,20 @@ export default class MentionCommand extends Command {
...
+			const removedLines: Array<string> = [];
+			for ( const line of lineArray ) {
+				if ( line.startsWith( '> ' ) ) {
+					removedLines.push( line );
+				}
+			}
+			for ( const remove of removedLines ) {
+				lineArray.splice( lineArray.indexOf( remove ), 1 );
+			}
+			messageText = lineArray.join( '\n' );
```
violine1101, line 46: This can be simplified a lot.
```suggestion
			messageText = messageText
				.split( '\n' )
				.filter( line => !line.startsWith( '> ' ) )
				.join( '\n' );
```

violine1101 (approved):

PR #199, opened 2021-05-12
Add list of possible commands to !jira help

chandler05:
## Purpose
Adds a list of possible bot commands to the !jira help message.

### DIFF
--- a/src/commands/HelpCommand.ts
+++ b/src/commands/HelpCommand.ts
@@ -13,11 +13,22 @@ export default class HelpCommand extends PrefixCommand {
 					Currently, the following projects are supported: ${ BotConfig.projects.join( ', ' ) }
 					To prevent the bot from linking a ticket, preface the ticket number with an exclamation mark.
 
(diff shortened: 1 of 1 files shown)


src/commands/HelpCommand.ts
```diff
@@ -13,11 +13,22 @@ export default class HelpCommand extends PrefixCommand {
...
+					This bot is continuously being worked on and this will receive more features in the future.
 					It is not possible to invite this bot to other servers yet.
 					If you have any issues, feel free to ping violine1101.
 
-					(For help with the bug tracker, use \`!jira tips\`)`.replace( /\t/g, '' ) )
+					(For help with the bug tracker or this Discord server, use \`!jira tips\`)`.replace( /\t/g, '' ) )
+				.addField( 'Bot Commands',
+					`\`!jira help\` - Sends this message.
+
+					\`!jira moo\` - Sends an embed with a specific (cow-themed) ticket.
```
violine1101, line 24: I personally don't think we should include this in the help. It's more like an easter egg.

chandler05: I think that is a good idea :)

violine1101 (approved):

PR #200, opened 2021-05-13
Limit daily requests for each user per channel

### LINKED ISSUE #166: Add request limit
## The Problem
We're currently very backlogged on requests and get more requests every day than we resolve. In order to solve that we need a request limit that doesn't forward requests from a user if that user has posted more than X requests in a single day.

## Possible Solution
Have a data base where pending requests are saved, and if there have been at least X requests by that user in that past 24 hours, deny that request.

This should be configurable on a per-channel basis.

chandler05:
## Purpose
Close #166
## Approach
Read from internal channels to get the requests that were created less than 24 hours before, limit per channel.

Negative request limits cause no limit for the channel.
## Future work
This and the MentionDeleteEventHandler (#201) both read from the user tag, so if it is changed then it will no longer match the user, there needs to be another way around this. I added the functionality to update the user tag and icon whenever a request is updated, but that doesn't entirely fix it.

### DIFF
--- a/config/template.yml
+++ b/config/template.yml
@@ -68,6 +68,16 @@ request:
     - <string>
     - ...
 
+  # The number of allowed requests per user per day.
+  # If the number of requests created by a single user exceeds this amount, new requests will not be forwarded to the internal channels.
+  # The length of this array MUST be the same as the length of `channels`.
+  # Setting to a negative number will remove the request limit
+  # Optional; empty by default.
+  requestLimits:
+    - <number>
+    - <number>
+    - ...
+
   # The IDs of the server's testing request channels.
   # Optional; empty by default.
   testingRequestChannels:
--- a/src/events/request/RequestEventHandler.ts
+++ b/src/events/request/RequestEventHandler.ts
@@ -90,7 +90,7 @@ export default class RequestEventHandler implements EventHandler<'message'> {
 
 		if ( requestLimit && requestLimit >= 0 && internalChannel instanceof TextChannel ) {
 			const internalChannelUserMessages = internalChannel.messages.cache
-				.filter( message => message.embeds[0].author.name == origin.author.tag )
+				.filter( message => message.embeds.length > 0 && message.embeds[0].author.name == origin.author.tag )
 				.filter( message => new Date().valueOf() - message.embeds[0].timestamp.valueOf() <= 86400000 );
 			if ( internalChannelUserMessages.size >= requestLimit ) {
 				try {
--- a/src/events/request/RequestEventHandler.ts
+++ b/src/events/request/RequestEventHandler.ts
@@ -100,7 +100,7 @@ export default class RequestEventHandler implements EventHandler<'message'> {
 				}
 
 				try {
-					const warning = await origin.channel.send( `${ origin.author }, you currently have created more requests than the daily limit for this channel, please wait until tomorrow before creating more requests.` );
+					const warning = await origin.channel.send( `${ origin.author }, you have posted a lot of requests today that are still pending. Please wait for these requests to be resolved before posting more.` );
 
 					const timeout = BotConfig.request.warningLifetime;
 					await warning.delete( { timeout } );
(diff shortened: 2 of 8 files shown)


dericksonmark (approved):

src/events/request/RequestEventHandler.ts
```diff
@@ -78,6 +84,33 @@ export default class RequestEventHandler implements EventHandler<'message'> {
...
+				.filter( message => new Date().valueOf() - message.embeds[0].timestamp.valueOf() <= 86400000 );
+			if ( internalChannelUserMessages.size >= requestLimit ) {
+				try {
+					await origin.react( BotConfig.request.invalidTicketEmoji );
+				} catch ( error ) {
+					this.logger.error( error );
+				}
+
+				try {
+					const warning = await origin.channel.send( `${ origin.author }, you currently have created more requests than the daily limit for this channel, please wait until tomorrow before creating more requests.` );
```
violine1101, line 103: This message might be misleading. If we resolve the requests quickly, the user won't need to wait until tomorrow.

I'd prefer a less technical message, although I'm not quite sure how that would look.

Perhaps something like `<username>, you posted a lot of requests today that are still pending. Please wait for these requests to be resolved before posting more.`

src/events/request/RequestEventHandler.ts
```diff
@@ -78,6 +84,33 @@ export default class RequestEventHandler implements EventHandler<'message'> {
...
 		}
 
+		const requestLimit = this.requestLimits.get( origin.channel.id );
+		const internalChannelId = this.internalChannels.get( origin.channel.id );
+		const internalChannel = await DiscordUtil.getChannel( internalChannelId );
+
+		if ( requestLimit && requestLimit >= 0 && internalChannel instanceof TextChannel ) {
+			const internalChannelUserMessages = internalChannel.messages.cache
+				.filter( message => message.embeds.length > 0 && message.embeds[0].author.name == origin.author.tag )
+				.filter( message => new Date().valueOf() - message.embeds[0].timestamp.valueOf() <= 86400000 );
```
violine1101, line 94: This line throws an error if `message.embeds[0]` is `undefined`. Which can happen if someone other than the bot writes in the channel. Actually we do have messages in our pending request channels at the very top with no embeds, which would cause the bot to crash as soon as a request is posted.

config/template.yml
```diff
@@ -68,6 +68,16 @@ request:
     - <string>
     - ...
 
+  # The number of allowed requests per user per day.
+  # If the number of requests created by a single user exceeds this amount, new requests will not be forwarded to the internal channels.
+  # The length of this array MUST be the same as the length of `channels`.
+  # Setting to a negative number will remove the request limit
+  # Optional; empty by default.
```
violine1101, line 75: Out of scope for this PR, but I don't like this invariant very much. We could definitely structure the request config a lot nicer.

I think I'll work on that since I'll be reworking some stuff regarding request channels anyway.

src/events/request/RequestEventHandler.ts
violine1101: Hm thinking about it actually the line above should prevent that... why doesn't it work?

violine1101: Ah, nvm. It crashes because `timestamp` is null, which is the case because I still have old requests without timestamps in my channel. Not sure if this is worth keeping in mind.

violine1101: Yeah, this needs to be changed. We still have timestamp-less messages in our pending requests channels.

violine1101 (changes requested):

src/events/request/RequestEventHandler.ts
violine1101: Actually nvm I'm wrong

src/events/request/RequestEventHandler.ts
chandler05: Ok, done. I think it looks fine.

violine1101 (approved):

Issue #202, opened 2021-05-13
Refresh internal requests on restart or add a command


chandler05:
## The Problem
When some changes are implemented to change the way internal request embeds look or add new reactions, old embeds keep the same properties and don't change.
## Possible Solution
These embeds could be readded on restart, or more likely using a command to prevent rate limits and other issues that may arise from readding them every time the bot restarts

violine1101:
I'd prefer an alternative solution, namely don't rely on Discord messages at all, but instead store all the stuff directly in a local data base that we can query without any delay. This would solve _so_ many issues we currently have, but I still haven't come around to add that data base part and figure out how to do it properly in JS/TS.

PR #203, opened 2021-05-13
Add/remove changed suggested emojis for internal channels on restart

chandler05:
## Purpose
On restart, refresh the new or removed suggested emojis for all internal channel messages.
## Notes
This does not slow down initialization, but I'm not sure if this is the worth having a feature for. I only added this since the suggested emojis are configurable and if the bot were usable on other servers and someone changed the emojis, they'd probably expect the old messages to change as well.

This potentially implements #202 if we don't want to refresh the embeds as well.

### DIFF
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -133,7 +133,7 @@ export default class MojiraBot {
 										}
 									} else {
 										if ( message.embeds.length == 1 && bot ) {
-											const reactedEmoji = reaction.emoji.toString();
+											const reactedEmoji = reaction.emoji.name;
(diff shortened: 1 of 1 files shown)


violine1101:
Adding emojis is painfully slow because Discord's rate limiting in this regard is ridiculous. When we add an emoji here, it could very well take the bot multiple hours to have added the new emojis, depending on the amount of existing pending requests. The alternative is just adding the emoji manually.

Not sure if it's really worth the hassle tbh.

Ultimately we can replace the reactions with the fancy new buttons, which will be even better anyway.

chandler05:
Buttons sound nice

Issue #98, opened 2020-08-08
Implement user verification process


violine1101:
## The Problem
There currently is no way to link a user's Discord account to their Mojira account. In some cases it's not obvious who which account belongs to, so a verification process for linking accounts between Discord and Mojira might be useful.

## Possible Solution
Add a `!verification <Mojira username>` command. When it is run, MojiraBot replies with a message like "To link your Discord account to your Mojira account "username", leave a comment on WEB-1234 with your Discord user ID (<user id>)."

MojiraBot also regularily checks that ticket WEB-1234 for newly incoming comments and checks them against the known Discord accounts. It then keeps all the linkings in a data base which can then be used to look up who which account belongs to, for example with `!jira who @<discord account>` or `!jira who <jira account>`.

Additionally, it might be nice to give verified users a special role.

Discussion needed, I'm not sure if this is really something that we want or need.

SPGoding:
If we decided to do this, it might be better that the user posted a random code generated by MojiraBot to that ticket instead of posting their Discord user ID

dericksonmark:
Hello! If this functionality added to MojiraBot, it might be better to delete the !verification command sent by the user immediately, and then send a private message to the command user with the response instead of just responding in the #bot-commands tab. If a malicious user tried to access Discord user IDs, sending private messages would prevent the user from being able to tell exactly which Discord account links to which Mojira account.

The verification code could expire within 10 minutes of it being sent and any record of it could be deleted from the database; this could help reduce the amount of data that would need to be stored in the database and help secure an account in case it was compromised.

I'm not quite sure what you meant by 'special role', but it should definitely not be too powerful. The only thing I can think of that would fit here would be a different colored name in Discord or Mojira. Adding affected versions to tickets or something like that is way too powerful. A malicious user could use this low-effort action to pretty much destroy the tracker.

violine1101:
> If this functionality added to MojiraBot, it might be better to delete the !verification command sent by the user immediately, and then send a private message to the command user with the response instead of just responding in the #bot-commands tab.

It would also be possible to only enable the !verification command in DMs, that way a user can only verify itself by writing a DM to MojiraBot.

The code being random was another thing I thought about, would definitely also be a good option to do it.

> I'm not quite sure what you meant by 'special role', but it should definitely not be too powerful.

I don't think it should make any difference permission-wise, I thought it should probably only be aesthetic (different color?), so that you know who has verified their account and who has not.

[...]

PR #209, opened 2021-05-28
Add a verification system

dericksonmark:
## Purpose
Add a verification system! Implements #98.
## Approach
To verify their account, a member will need to:
1. Send `!jira verify` (MojiraBot will send them a 15 character token)
2. Comment on a specific Mojira ticket with their token
3. Send a private message to the bot with `link <jira username>`

The bot will check the ticket for their comment, and if the username and code matches, the accounts will be linked.

This also adds:
- A command `!jira who` or `!jira whois` to check the account linked to either a Mojira or a Discord account.
- A method to remove the verification role from a user temporarily, by reacting to the log message.
- A method to remove the verification link from a user, by reacting to the log message with ✂.

## Future work


src/commands/WhoisCommand.ts
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+
+		const logChannel = await DiscordUtil.getChannel( BotConfig.verification.verificationLogChannel );
+		const allMessages: Message[] = [];
+		let lastId: string | undefined;
+		let continueSearch = true;
+
+		try {
+			while ( continueSearch ) {
```
chandler05, line 37: You're going to want to move this to `MojiraBot.ts` because we don't want to be fetching all messages every time the command is run. All we need to do is fetch these on startup and read from `logChannel.messages.cache`.

dericksonmark: That is a very good point, thank you!

src/MojiraBot.ts
```diff
@@ -86,6 +86,30 @@ export default class MojiraBot {
...
+							options.before = lastId;
+						}
+						const messages = await pendingChannel.messages.fetch( options );
+						allMessages.push( ...messages.array() );
+						lastId = messages.last()?.id;
+						if ( messages.size !== 50 || !lastId ) {
+							continueSearch = false;
+						}
+					}
+					this.logger.info( `Fetched ${ allMessages.length } messages from "${ pendingChannel.name }"` );
```
urielsalis, line 109: Can it delete the messages when they are verified or it expires? to save memory here

src/commands/VerifyCommand.ts
```diff
@@ -0,0 +1,101 @@
...
+			if ( foundUser ) {
+				await message.channel.send( `${ message.author }, you already have a pending verification request!` );
+				await message.react( '❌' );
+				return false;
+			}
+
+		}
+
+		try {
+			const token = this.randomString( 15, '23456789abcdeghjkmnpqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ' );
```
urielsalis, line 54: Token should be something like MOJIRA-DISCORD-VERIFICATION-<random>

Just to make sure people dont trick users into doing it

src/MojiraBot.ts
dericksonmark: I have just made changes to this. Have these changes fixed the issue here?

dericksonmark:
I think this is as complete as I'm going to get it, so I am marking this pull request as ready for review.

src/commands/VerifyCommand.ts
```diff
@@ -0,0 +1,116 @@
...
+			if ( foundUser ) {
+				await message.channel.send( `${ message.author }, you already have a pending verification request!` );
+				await message.react( '❌' );
+				return false;
+			}
+
+		}
+
+		try {
+			const token = 'DISCORD-VERIFICATION-' + this.randomString( 15, '23456789abcdeghjkmnpqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ' );
```
Marcono1234, line 68: Might be good to add a comment, explaining why some chars (`0`, `o`, `O`; `1`, `I`) are omitted.
And why are `f` and `i` missing? With which characters can they be confused?

dericksonmark: You're definitely right. Good idea, thank you!
I don't know why I removed `f`. It's very possible that it was just a mistake.
I think I removed `i` because it could be confused with `I`, but now I realize that they look different enough in Discord that they wouldn't be confused.

chandler05:
Looks like the team is in favor of adding this according to https://discord.com/channels/647810384031645728/647810521713606656/850795026674089985

dericksonmark:
I've decided to close this for now. When creating this, I made many choices that are inefficient for the bot, like storing verifications in a Discord channel. After the jira.js and discord.js pull requests are merged, I plan to remake this in a more efficient way. :)

PR #214, opened 2021-06-02
Shorten Jira error messages

### LINKED ISSUE #212: Improve error logging for jira connection errors
## The Problem
Currently the log is spammed with huge amounts of error info when the bot cannot connect to JIRA.

## Possible Solution
Only log relevant information (i.e. status code, message if available).

dericksonmark:
## Purpose
Fixes #212.
## Approach
This changes the error message in Task.ts, FilterFeedTask.ts, and VersionFeedTask.ts to be only four lines long. These four lines include the status code, status text, and error messages (some of these may not be defined).

These error messages used to be over 300 lines each.

The changes here don't affect other Jira error messages besides these three, but these are the errors that are causing the most spam when the bot can't connect to Jira.

Here is an example of a shortened error message.

Screen Shot 2021-06-01 at 11 35 53 PM


violine1101 (commented):
It might be better to extract this into a common util function that stringifies a JIRA error, otherwise we have duplicated code here which is not good.

dericksonmark:
I think I've added a LoggerUtil properly, so I'll mark this pull request as ready for review.

src/util/LoggerUtil.ts
```diff
@@ -0,0 +1,9 @@
+export class LoggerUtil {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public static shortenJiraError( error: any ): string {
+		return `Information:
+		  Status code: ${ error.response?.status || undefined }
+		  Status text: ${ error.response?.statusText || undefined }
+		  Error message(s): ${ error.response?.data.errorMessages || undefined }`.replace( /\t/g, '' );
```
SPGoding, line 7: Potential NPE could occur here. I also don't see the need for `|| undefined` as it could replace 0 or empty strings with `undefined`.

```suggestion
		  Status code: ${ error?.response?.status }
		  Status text: ${ error?.response?.statusText }
		  Error message(s): ${ error?.response?.data?.errorMessages }`.replace( /\t/g, '' );
```

dericksonmark: Thanks! Sorry, I'm not sure how I missed those.

PR #215, opened 2021-06-02
Do not create a mention embed if the message contains a certain keyword

dericksonmark:
## Purpose
Adds #14.
## Approach
If `forbiddenTicketText` (default is `!nm` [no mention]) is set in `default.yml`, MojiraBot won't create a ticket embed if the message contains the forbidden text.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -10,6 +10,8 @@ quotedTicketsCauseEmbed: false
 
 forbiddenTicketPrefix: '!'
 
+forbiddenTicketText: '!nm'
+
 requiredTicketPrefix: ''
 
 embedDeletionEmoji: '🗑️'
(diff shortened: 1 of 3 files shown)


violine1101:
To provide a bit of context, I've created this issue because the way in which MojiraBot creates ticket embeds is somewhat obtrusive (only mentioning a ticket suffices to trigger the bot). Thus I explored some solutions to this problem, one of which was to have some keyword that disables mentions.

However, I think by now (1.5 years later) I think the better option is just to only trigger the bot if the ticket ID is prefixed with `!`, like how most bots do it. That's just way more user friendly.

In this case, this feature would just not be needed at all anymroe (which would be good because I don't really like MojiraBot becoming more and more complex)

Wanted to bring this up in Discord already, but haven't come around to it.

violine1101:
Perhaps in general, in the beginning of the project I used the issues here as an idea dump and not as "something that we need". I should go through all of that and sort out what is actionable and what isn't.

PR #220, opened 2021-06-12
Add .DS_Store files to .gitignore

dericksonmark:
## Purpose
Ignore .DS_Store files. These are files that store folder data on macOS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ node_modules
 
 .vscode/*
 .history
-.DS_Store
+**/.DS_Store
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -3,8 +3,9 @@ node_modules
 
 .vscode/*
 .history
+.DS_Store
```
violine1101, line 6:
```suggestion
**/.DS_Store
```
This ignores all `.DS_Store` files in all folders

violine1101 (approved):

PR #223, opened 2021-07-09
Use crosspost from discord.js

SPGoding:
## Purpose
So that we could get better rate limit handling :eyes:


violine1101 (commented):
Two things:

* You removed the error handling when trying to crosspost a message. Not a big deal since the only thing that wouldn't be done if crossposting fails is that it doesn't react with the emoji, but at least I think the correct error message should be sent. (TBH I'm not a huge fan of exceptions and catching them all the time is really annoying and I wish it was easier to do proper error handling in TS but it is how it is ¯\\\_(ツ)_/¯ )
* I think because of this change `BotConfig.token` could now be made private.

violine1101 (approved):

PR #225, opened 2021-07-10
Reject exact duplicate requests

### LINKED ISSUE #222: Merge or delete duplicate requests
## The Problem
Duplicate requests happen often

## Possible Solution
If an exact duplicate of another request is requested, we either merge them into a single internal request or reject the duplicate request

chandler05:
## Purpose
Fix #222 
## Approach
Reject any duplicate requests after searching through pending for exact matches


violine1101:
As I already said in #222, I don't think this is something we should implement.

PR #227, opened 2021-07-11
Modmail

### LINKED ISSUE #226: Forward direct messages sent to the bot to an internal channel
## The Problem
Occasionally, issues containing personal information are reported publicly on the bug tracker, and moderators and helpers don't check that ticket within a reasonable amount of time. This allows for the personal information contained in that report to be available for much longer than it should be. Currently, the Moderator ping is often used to inform moderators when a report is created like this, but this creates many issues (others can see the report linked in the message, moderators are pinged for non-urgent cases, etc).
## Possible Solution
Allow users to privately message MojiraBot with a request to set an issue to private (can also be used for other requests that shouldn't be done through public communication). This can be forwarded to a private channel, and then helpers/mods/Mojang can reply to specific messages. All of this would have to be done without interrupting private messages sent in verification (#209), if that is implemented.

chandler05:
## Purpose
Resolve #226 (probably worth discussing before merging, but it has already been discussed and the general consensus is so far in favor of this)
Add the database functionality!
## Approach
DMs to Mojira Bot are forwarded to an internal channel and put into a thread.
/modmailban to ban a user from using the modmail system.
/modmailunban to unban a user from using the modmail system.
Reply in a modmail message thread to send a reply back to the user.
Threads will be archived after 24h of inactivity, after which a new thread will be created for subsequent messages
## Future work
New channel needed for modmail, combine with #209's message event handler.

~Add direct reply functionality when discord.js v13 is added.~


chandler05:
Looks like the `package-lock.json` file got a bunch added to it when I ran `npm install`. I'm not sure which of it should be kept or removed, if any.

src/events/modmail/ModmailEventHandler.ts
```diff
@@ -0,0 +1,38 @@
...
+				await modmailChannel.send( `${ origin.author }: ${ origin.content }` );
+				await origin.react( '📬' );
+			} catch ( e ) {
+				this.logger.error( e );
+
+				return;
+			}
+		} else if ( banStatus !== undefined ) {
+			try {
+				await origin.channel.send( 'We\'re sorry, but you have been banned from sending any further modmail messages.' );
```
chandler05, line 29: Since we haven't taken DM rate limiting into account, this could be removed and left with the reaction to the message instead.

dericksonmark:
Could this also forward screenshots/images attached to the message? A user may attach one to a modmail message.

violine1101:
I haven't fully looked at this PR but it's certainly something I'm excited about.

With the new Discord features now available, I think the best solution would be to have a thread for each user who messages MojiraBot via DMs in the modmail channel. That's probably not too easy to implement though and as such out of scope for this PR, so let's just land this first and then we can still expand on it later if necessary. (tbh I don't think the volume of modmail would be too large anyway)

chandler05:
I'll need to refactor parts of the PR to match discord's new changes, so I can add threads while I do that :)

chandler05:
I've switched the system to use threads, and it will now forward attachments on messages as well. Threads are archived after 24 hours, so a new thread will be created for new messages from that user afterwards, just to make sure modmails are seen. This one is now ready for review!

violine1101 (approved):
I'd like to apologize for taking this long to look at the PR. I don't have a good reason for that, I just procrastinated on it too much.

Thank you for adding this - it's very much needed. No concerns about the code itself.

I had concerns over the DM channel no longer being usable as a personal place to try out commands, but I noticed that with the slash commands this is not really possible anymore anyway.

I do not like MojiraBot having too many distinct feature sets: It is responsible for ticket embeds, feeds, requests, and now also modmail. Optimally I'd like to separate these four responsibilities into four separate bots for easier maintenance. This would also allow us to finally make this bot publicly accessible, in particular the parts of it responsible for ticket embeds (and maybe also feeds). But this is way out of scope for this PR.

With that being said, let's finally merge this!

PR #229, opened 2021-07-15
Add channel configuration options using the channel's topic

### LINKED ISSUE #2: Channel configuration
It would be good if the bot's actions could be customized depending on the channel. For instance, we might only want a limited issue preview in the request channels on our Discord server.

chandler05:
## Purpose
Resolves #2 
## Approach
Uses specific tags in the channel's topic to indicate different config options

Current tags are:
`~limited-info` - shows only the ticket's title and description in embeds
`~no-mention` - disables mentions
`~no-command` - disables commands
`~public-search` - makes /search results public
## Future work
I'd be open to adding other config options, but currently I have no ideas. Feel free to let me know if there are any other config options that would be good to have! :)


src/events/message/MessageEventHandler.ts
```diff
@@ -42,21 +42,12 @@ export default class MessageEventHandler implements EventHandler<'messageCreate'
...
-
-			// We want the bot to create embeds in testing channels if someone only posts only a ticket ID
-			// so that people know what the issue is about
 		} else if ( BotConfig.request.internalChannels && BotConfig.request.internalChannels.includes( message.channel.id ) ) {
 			// This message is in an internal channel
 			await this.internalProgressEventHandler.onEvent( message );
-
-			// Don't reply in internal request channels
-			return;
 		}
```
violine1101, line 51: I think we should still keep these early returns, just so we don't have to add the configuration options to all the channel descriptions

src/commands/BugCommand.ts
```diff
@@ -15,6 +16,10 @@ export default class BugCommand extends SlashCommand {
 		);
 
 	public async run( interaction: ChatInputCommandInteraction ): Promise<boolean> {
+		if ( interaction.channel === null ) return false;
+
+		if ( ChannelConfigUtil.mentionsDisabled( interaction.channel ) ) return false;
```
violine1101, line 21: This doesn't seem right, wouldn't this rather be covered by the ~no-command flag?

violine1101 (commented):
Two comments, apart from that this should be good to go!

src/commands/BugCommand.ts
chandler05: 👍

src/events/message/MessageEventHandler.ts
chandler05: Good Point. I think I did that originally to allow more configuration, but I do think that its unecessary

violine1101 (approved):

Issue #231, opened 2021-08-31
Support logging into Mojira


dericksonmark:
## The Problem
During testing, a user may want to have MojiraBot log into an account so that it can access private testing tickets.
## Possible Solution
Allow MojiraBot to be configured in a way that allows it to log into the bug tracker.

PR #232, opened 2021-08-31
Allow MojiraBot to log into Mojira

dericksonmark:
## Purpose
Implement #231.
## Approach
Adds `jiraUsername` and `jiraPassword` options in the `local.template.yml` (optional; empty by default).


dericksonmark:
I'm closing this because personal access token-based authentication would probably be better in this case.

PR #237, opened 2021-09-22
Transfer replies from command message

dericksonmark:
## Purpose
Have MojiraBot transfer replies from the command message to the message it sends.
I don't think this can or should be merged until #236 is merged.
## Approach
Creates the `DiscordUtil.sendMentionMessage` method, and switches from `message.channel.send` in MentionCommand, TipsCommand, and HelpCommand to `DiscordUtil.sendMentionMessage`.


violine1101 (approved):
With slash commands being implemented in #263, this now is only relevant for mention commands. I'll merge this first and then the slash commands PR afterwards, which will only keep this logic for mentions.

Issue #238, opened 2021-09-22
Add more features to !jira search


chandler05:
## The Problem
Currently the !jira search command is a bit lacking in features
## Possible Solution
Add new options for different search parameters such as project and resolutions.

PR #241, opened 2021-09-22
Initial support for search arguments

dericksonmark:
## Purpose
#238 should be fixed by this.

Adds arguments to the !jira search command!

There are two types of arguments that can be used with this, which are arguments that start with `:` and arguments that start with `_`.

Arguments starting with `:` are single-word arguments, and there's currently only one of these: `:query`. This argument allows users to use a JQL query that they write themself, instead of using the building blocks that the !jira search command provides. Example usage: `!jira search :query text ~ crash AND reporter = markderickson`

Arguments starting with `_` are multi-word arguments, and there are lots of these: `_project`, `_creator`, `_reporter`, `_assignee`, `_version`, `_fixversion`, `_resolution`, `_status`, `_confirmation`, `_gamemode`, `_id`, `_key`, `_labels`, `_priority`, `_mp`, `_comment`, `_summary`, `_description`, `_environment`. Each one of these can be paired with a corresponding value to build onto the query. Using only multi-word arguments allows for usage of the command as it works currently. Example usage: `!jira search _project MC _labels crash _reporter markderickson _mp "Very Important" _confirmation "Community Consensus" height limit crash 256`

I will work on this once the bot is updated to discord.js version 13!


dericksonmark:
Closed in favor of #261.

PR #244, opened 2021-09-22
Update to discord.js v13 (third try)

violine1101:
Previous PRs:
#224, #236

Reverted in:
#235, #243

This change requires more testing and changes to the codebase to ensure it works properly with discord.js v13.

### DIFF
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -49,10 +49,10 @@ export default class MojiraBot {
 	private static running = false;
 	private static botUser: ClientUser;
 
+	// TODO: integrate newErrorHandling from Jira.js
 	public static jira = new JiraClient( {
 		host: 'https://bugs.mojang.com',
 		telemetry: false,
-		newErrorHandling: true,
 	} );
 
 	public static async start(): Promise<void> {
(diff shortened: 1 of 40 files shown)


mdashlw:
What's the status of this?

violine1101:
Essentially, upgrading to discordjs v13 broke stuff and we don't know why it broke stuff or how to fix it.

I'm currently in the process of rebasing everything, and then I'll merge this after having tested it locally.

chandler05:
I noticed that reactions to the pending request seem to appear slower when a request is added. Is this just my bot or does it happen for yours as well?

chandler05 (approved):

violine1101:
I'm not sure. It's always a little delayed, however possibly the delay has been increased with v13. I think for our purposes it's fine.

In the future it'd be nice to replace this with buttons anyway, but we aren't that far yet.

src/MojiraBot.ts
```diff
@@ -52,6 +52,7 @@ export default class MojiraBot {
 	public static jira = new JiraClient( {
 		host: 'https://bugs.mojang.com',
 		telemetry: false,
+		newErrorHandling: true
```
violine1101: I looked into changing this but feared that it might mess up error reporting.

If possible we should probably do this in a separate PR.

violine1101:
Forgot to include co-authors in the commit, new commit is 20ee09e08777f826b3c708a69bcff814e8911e34 (no other changes)

PR #252, opened 2021-11-18
Update jira.js to v2.8.x and fix some null/style issues

violine1101:
Just updating jira.js, and doing some minor tweaks along the way.
I'd recommend to view each commit separately.


src/events/mention/MentionDeleteEventHandler.ts
```diff
@@ -10,15 +10,8 @@ export default class MentionDeleteEventHandler implements EventHandler<'messageR
...
-
-		try {
-			const embeds = message.embeds;
-			const mentionEmbed = embeds[0];
-			footer = mentionEmbed.footer.text;
-		} catch ( error ) {
-			this.logger.error( error );
-		}
+		const footer = message.embeds[0]?.footer?.text;
+		if ( footer !== undefined ) return;
```
dericksonmark, line 14: This should use `===` instead of `!==`. Currently, the code following this will only run if the footer is undefined, which would cause problems.

Jack-McKalling: Agreed. Simple overlook

src/tasks/VersionFeedTask.ts
```diff
@@ -220,7 +222,7 @@ export default class VersionFeedTask extends Task {
 		};
 
 		try {
-			versionIssueCounts = await MojiraBot.jira.projectVersions.getVersionsRelatedIssuesCount( {
+			versionIssueCounts = await MojiraBot.jira.projectVersions.getVersionRelatedIssues( {
```
dericksonmark, line 225: The documentation for jira.js seems to imply that this method also returns the usage of this version in custom fields. Would this allow it to continue to comply with the type defined above?
Screen Shot 2022-01-07 at 11 35 24 PM

dericksonmark (approved):

MrRefactoring (approved):

PR #257, opened 2022-02-06
Properly format links surrounded with angle brackets

dericksonmark:
## Purpose
Links in requests are not properly formatted in requests when surrounded by angle brackets.
## Approach
Add `<` and `>` to the regular expression to match a link in a request, and then remove them when returning it. I think this should work.

### DIFF
--- a/src/util/RequestsUtil.ts
+++ b/src/util/RequestsUtil.ts
@@ -98,7 +98,7 @@ export class RequestsUtil {
 	 * @returns A NEW regex object every time. You have to store it as a variable if you use `exec` on it, otherwise you will encounter infinite loops.
 	 */
 	public static getTicketRequestRegex(): RegExp {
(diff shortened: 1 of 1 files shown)


src/util/RequestsUtil.ts
```diff
@@ -141,8 +141,8 @@ export class RequestsUtil {
 	public static replaceTicketReferencesWithRichLinks( content: string ): string {
 		const regex = new RegExp( `${ this.getTicketRequestRegex().source }(?<query>\\?[^\\s#]+)?(?<anchor>#\\S+)?`, 'g' );
 
-		// Escape all of the following characters with a backslash: [, ], \
-		return content.replace( /([[\]\\])/gm, '\\$1' )
+		// Escape all of the following characters with a backslash: [, ], \, <, >
+		return content.replace( /([[\]\\]<>)/gm, '\\$1' )
```
SPGoding, line 145:
```suggestion
		return content.replace( /([[\]\\<>])/gm, '\\$1' )
```

dericksonmark: Oops, sorry. Thanks!

SPGoding (approved):

PR #258, opened 2022-02-13
Fix regular expressions for requests with angle brackets

dericksonmark:
Thank you @SPGoding for a lot of help with this. :D

## Purpose
(Hopefully) fix the problems caused by #257.


SPGoding (approved):
Thanks!

PR #263, opened 2022-05-14
Implement slash commands

### LINKED ISSUE #181: Use Discord's built-in slash commands
Discord recently added slash commands which provide a nice interface for triggering commands with a bot. We should investigate this and see how we can implement that for MojiraBot.

chandler05:
## Purpose
Fix #181

## Approach
Convert all existing commands over to slash commands, with minor changes in syntax and behavior.
I did my best to keep the new commands as similar to the old commands as possible.

## Future work
Clean up my somewhat messy code


config/main.yml
```diff
@@ -8,6 +8,8 @@ owners:
   - '417403221863301130' # violine1101
   - '437088929485684737' # SPGoding
 
+guild: '647810384031645728' # Mojira
```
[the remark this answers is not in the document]
chandler05: Good point. I have removed this from the config and instead it now registers the commands for all guilds that the bot is currently in.

src/commands/HelpCommand.ts
```diff
@@ -1,11 +1,13 @@
...
+export default class HelpCommand extends SlashCommand {
+	public readonly slashCommandBuilder = this.slashCommandBuilder
+		.setName( 'jira-help' )
+		.setDescription( 'Sends a help message.' );
 
-	public async run( message: Message ): Promise<boolean> {
+	public async run( interaction: ChatInputCommandInteraction ): Promise<boolean> {
 		try {
 			const embed = new EmbedBuilder();
 			embed.setTitle( '<:mojira:821162280905211964> **MojiraBot help** <:mojira:821162280905211964>' )
```
violine1101, line 13: With slash commands, the bot appears to have the same restrictions for emojis as regular users. For us that's okay but it becomes a slight issue as soon as we let the bot to be invited to other servers. However when we do that we'll probably rework this command anyway.

src/commands/BugCommand.ts
```diff
@@ -30,39 +40,26 @@ export default class BugCommand extends PrefixCommand {
...
 		if ( embed === undefined ) return false;
 
-		embed.setFooter( { text: message.author.tag, iconURL: message.author.avatarURL() ?? undefined } )
-			.setTimestamp( message.createdAt );
+		embed.setFooter( { text: interaction.user.tag, iconURL: interaction.user.avatarURL() ?? undefined } )
+			.setTimestamp( interaction.createdAt );
 
 		try {
-			await message.channel.send( { embeds: [embed] } );
+			await interaction.reply( { embeds: [embed], ephemeral: true } );
```
violine1101, line 57: I think for this command it makes more sense to have the command reply public rather than ephermal.

src/commands/BugCommand.ts
```diff
@@ -30,39 +40,26 @@ export default class BugCommand extends PrefixCommand {
 			embed = await mention.getEmbed();
 		} catch ( err ) {
 			try {
-				await message.channel.send( err );
+				await interaction.reply( { content: err, ephemeral: true } );
```
violine1101, line 43: The `err` variable we get here usually does not result in an error message that can just be sent as a reply. This results in the broad error message "An error occurred while running this command" always if JIRA returns an error. E.g. private bug report, too large ticket number, deleted bug report, etc.
This might be out of scope for this PR since error handling with MojiraBot is a mess everywhere at the moment. Might be good for a follow-up PR though.

src/commands/commandHandlers/SlashCommandRegister.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Routes } from 'discord-api-types/v9';
+import { Client, Collection, ChatInputCommandInteraction } from 'discord.js';
+
+export default class SlashCommandRegister {
+	public static async registerCommands( client: Client, token: string ) {
+		client.guilds.cache.forEach( async fetchedGuild => {
+			await fetchedGuild.fetch();
+
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
+			const commands: any[] = [];
```
violine1101, line 13: I dislike the fact that this is untyped, is there no better option?

violine1101 (approved):
Great work!

There still are a few rough edges though, but I'd rather merge this PR now since it's so large, and then sandpaper over it later.
For example, our notion of "mentions" could be completely separated from the command system into its own thing as it's the only thing left using the old commands system. This would probably clean up the code a bit more. But, as I said, out of scope for this PR.

src/commands/commandHandlers/SlashCommandRegister.ts
chandler05: Yes, it can be typed. It just took a bit of digging to find what to type it to.

PR #265, opened 2022-05-17
Remove await from message publishing, add previous fixes to cached filter feed

chandler05:
## Purpose
Fix infinite posts in the fi feeds while waiting for the rate limit to expire

### DIFF
--- a/src/tasks/FilterFeedTask.ts
+++ b/src/tasks/FilterFeedTask.ts
@@ -78,7 +78,9 @@ export default class FilterFeedTask extends Task {
 				}
 
 				if ( this.publish ) {
-					NewsUtil.publishMessage( filterFeedMessage );
+					NewsUtil.publishMessage( filterFeedMessage ).catch( err => {
+						FilterFeedTask.logger.error( `[${ this.id }] Error when publishing message`, err );
+					} );
 				}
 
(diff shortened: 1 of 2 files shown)


src/tasks/FilterFeedTask.ts
```diff
@@ -78,7 +78,9 @@ export default class FilterFeedTask extends Task {
 				}
 
 				if ( this.publish ) {
-					await NewsUtil.publishMessage( filterFeedMessage );
+					NewsUtil.publishMessage( filterFeedMessage ).catch( err => {
```
violine1101, line 81: In theory the catch is not necessary since `NewsUtil.publishMessage` never throws

violine1101: But I guess it's ESLint that then complains?

violine1101: Ah right. I see. I guess that's fine then.

violine1101 (approved):

src/tasks/FilterFeedTask.ts
chandler05: Yeah I originally just removed the await until ESLint wanted me to do that as an alternative.

PR #280, opened 2023-06-13
Add support for pomelos

randomairborne:
## Purpose
`#0` looks silly and is inaccurate

## Approach
just check the discrim

## Future work
When discord.js fixes the `tag` field, this could be removed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "mojira-discord-bot",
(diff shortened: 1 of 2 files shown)


violine1101 (approved):
Looks good, but it's more of a temporary solution. When the rollout is complete we'll probably upgrade to a new discord.js version anyway and then we can just remove the discriminator everywhere.

randomairborne:
unfortunately the libraries seem to be reacting slowly to this, which is a shame. Thanks.

Issue #201, opened 2021-05-13
Username changes cause mention deletions not to work properly


chandler05:
## The Bug
Because user's can change their name and tag, the mention deletion does not always work because it tries to match the current user's tag to the one in the footer which doesn't update.
## Expected behavior
The user can delete mention embeds even after changing their name.

PR #281, opened 2023-06-25
Display user names the same everywhere

violine1101:
## Purpose
Display new user handles as `@violine1101` and old ones as `violine1101#8993` everywhere.

This also should fix the issue where messages could not be deleted if you had a new user handle.

## Approach
Add a helper function in `DiscordUtil` to decide how to print the name. Also added two more helper function for consistent setting of embed footers and authors.

## Future work
We'll probably need to rework this with the next discord.js update anyway, I presume (though I haven't checked their progress on the matter)

### DIFF
--- a/src/MojiraBot.ts
+++ b/src/MojiraBot.ts
@@ -91,7 +91,9 @@ export default class MojiraBot {
 
 			this.botUser = this.client.user;
 			this.running = true;
-			this.logger.info( `MojiraBot has been started successfully. Logged in as ${ this.botUser.tag }` );
+			this.logger.info(
+				`MojiraBot has been started successfully. Logged in as ${ DiscordUtil.getUserHandle( this.botUser ) }`
+			);
 
 			// Register events.
 			EventRegistry.setClient( this.client );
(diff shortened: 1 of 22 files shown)


chandler05:
I'll take a look at this at some point, but from what I can tell this would still run into the issue of #201. As far as I'm aware, using the user ID is the only way to identify the user correctly rather than the username. Not sure if you'd want to try to address this here and now or if you'd rather do that later.

chandler05:
I can take a look at this later, but from what I can tell this would still leave #201 as an issue. Not sure if you'd want to figure out a solution for that here, or if you'd rather leave that alone right now.

violine1101:
I've specifically tested that case and it should work (though only with my second commit)

mdashlw:
Missed this: https://github.com/mojira/mojira-discord-bot/blob/437f7b51ad8a76fbf5035385804f82d69eef6a27/src/util/RequestsUtil.ts#L75

---
discord.js support for pomelo was added in 14.12.0 (currently using 14.2.0). `User#tag` now returns username or username#tag whereas this getUserHandle returns `@username`

PR #299, opened 2025-01-21
Publish version feed messages

dericksonmark:
## Purpose
Publish version feed messages, which are currently sent but not distributed to other servers.


violine1101:
note: there's a limit on 5 published messages per hour which is why i initially didn't enable this feature but i suppose we can deal with that when it comes to that

PR #300, opened 2025-02-21
Use 'bugs-legacy' api route

lowercasebtw:
## Purpose

As mojira is transitioning to a new format, and their current route is down, use the legacy route which is still working for now till the new API becomes public.

## Approach

Modified src/BotConfig.ts 'host' to be 'bugs-legacy.mojang.com'

## Future work

?


JarlPenguin:
Superseded by #303

PR #314, opened 2025-08-02
Update MentionDeleteEventHandler to reflect Discord's new username format

vcosm:
## Purpose
Update MentionDeleteEventHandler to reflect Discord's new username format, as it currently looks for legacy _usernames#1234_ that have been phased out
## Approach
New regex pattern that follows Discord's new username format


Greymagic27:
Thank you :)

